PR #1, opened 2020-09-30
WIP: re-implement the current Ion experience

hamilton:
This PR is still in-progress but wanted to share this as a way to put together a to-do list of things to be taken care of before this can land:

- [x] __refactor the state abstractions__ – at the moment, I'm using a pretty blunt localStorage abstraction that implements a `setItem` and `getItem`. But it doesn't account for message passing that needs to happen when certain events fire (such as enrolling). This should be straightforward to mock up and implement. To my knowledge, we primarily need to abstract out `updateIonEnrollment`, `updateStudyEnrollment`, and an API for `getAvailableStudies`.
- [x] __mock up available studies in the prototype web version__ – the available studies should be a separate "request" in the API.
- [x] __reduce the CSS down to a smaller subset__ – we currently have `common.css` & `pioneer.css`, both of which I copied from the current implementation, loading in the html file. We probably don't need both of these.
- [x] __move the copy components into their own directory away from `/src/components`__
- [x] ~__validate the UX for enrolling / leaving a study__ – should there be a modal there?~ (will have this be in a follow-on PR)
- [x] __cover bases for `about:ion`__ – I think this is our first target for this app, so I want to make sure all our bases are covered.
- [x] __clean up final pass__

More to come.

ion-web-01


rhelmer:
This is looking great, thanks for working on this! More below:

> This PR is still in-progress but wanted to share this as a way to put together a to-do list of things to be taken care of before this can land:
> 
>     * [ ]  **refactor the state abstractions** – at the moment, I'm using a pretty blunt localStorage abstraction that implements a `setItem` and `getItem`. But it doesn't account for message passing that needs to happen when certain events fire (such as enrolling). This should be straightforward to mock up and implement. To my knowledge, we primarily need to abstract out `updateIonEnrollment`, `updateStudyEnrollment`, and an API for `getAvailableStudies`.

I think the actions we need to support are:
- set un/enroll from Ion itself
- set un/enroll from individual studies
  - keep in mind that studies may not necessarily be add-ons!
- get all available studies
- get active studies the user is enrolled in

It's unclear to me exactly what the `update` functions above would do, if they (for instance) took an optional argument and otherwise returned the current status that would be enough (I don't know what is idiomatic for Svelte apps, I'd imagine these as separate getters/setters from my naive vanilla JS PoV :) )

I do really like the localstore implementation, should make it much easier for automated tests and to hack on the design locally. This will only need minor changes to switch to webextension storage for instance.

> 
>     * [x]  **mock up available studies in the prototype web version** – the available studies should be a separate "request" in the API.
> 
>     * [ ]  **reduce the CSS down to a smaller subset** – we currently have `common.css` & `pioneer.css`, both of which I copied from the current implementation, loading in the html file. We probably don't need both of these.
> 
>     * [ ]  **move the copy components into their own directory away from `/src/components`**
> 
>     * [ ]  **validate the UX for enrolling / leaving a study** – should there be a modal there?

It's really easy to accidentally unenroll (which deletes your data and the study may not be re-enrollable) so I suspect we want *something* there but happy to defer to UX :)

>     * [ ]  **cover bases for `about:ion`** – I think this is our first target for this app, so I want to make sure all our bases are covered.
> 
>     * [ ]  **clean up final pass**
> 
> 
> More to come.
> 
> ion-web-01

Again this is all great, thanks for taking this on!

hamilton:
@rhelmer I made the above changes, I think within reason. I would be comfortable with us merging this so we can get this into an add-on.

src/StudyCard.svelte
```diff
@@ -1,66 +1,124 @@
...
   <div class="icon">
     <slot name="icon"><span class="missing" /></slot>
   </div>
 
   <div class="description">
     <slot name="description"><span class="description" /></slot>
+    <p>
+      You can always find the
+      <a
+        href="https://addons.mozilla.org/en-US/firefox/addon/pioneer-v2-example/">privacy
```
rhelmer, line 109: This should come from the study metadata I think?

rhelmer (approved):
I made a note but I think this is so much better than what is on the default branch we should just merge it now.

src/StudyCard.svelte
hamilton: thanks for catching, will file an issue to track this

hamilton: Tracked in #4

PR #7, opened 2020-10-02
Add test coverage for the Ion Core addon

Dexterp37:
This additionally moves the Ion Core code to its own directory, adds test coverage support and a rollup.js configuration.

### DIFF
--- a/core-addon/IonCore.js
+++ b/core-addon/IonCore.js
@@ -1,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = class IonCore {
+  initialize() {
+    // Whenever the addon icon is clicked, open the control page.
+    chrome.browserAction.onClicked.addListener(this._openControlPanel);
+    // After installing the addon, make sure to show the control page.
+    chrome.runtime.onInstalled.addListener(this._openControlPanel);
+  }
+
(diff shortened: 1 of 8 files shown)


core-addon/IonCore.js
```diff
@@ -0,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+module.exports = class IonCore {
+  initialize() {
```
rhelmer, line 6: Any reason not to make this `constructor` so it's not necessary for the caller to call this method too?

manifest.json
```diff
@@ -32,7 +32,10 @@
   },
 
   "background": {
-    "scripts": ["background.js"]
+    "scripts": [
+      "node_modules/webextension-polyfill/dist/browser-polyfill.min.js",
```
rhelmer, line 36: Good idea :) The promise-based `browser.*` API is so much nicer, will be good to not have to hamstring ourselves to `chrome.*`.

rhelmer (approved):
Just a question on `initialize` method vs. `constructor`, lgtm.

core-addon/IonCore.js
Dexterp37: Only testing convenience! Happy to change this moving forward as required.

manifest.json
Dexterp37: Ouch! I just noticed I'm including this but still using `chrome` instead of `browser`. Let me fix this in a follow-up PR

PR #11, opened 2020-10-02
adds web-extension.js API endpoint for frontend; web extension now pulls available studies from Remote Settings

### LINKED ISSUE #2: add web extension API implementation
Once #1 merges, we'll need to make sure the web extension API endpoint in `/src/api/web-extension.js` is implemented.

### LINKED ISSUE #10: add-on needs to fetch available studies from remote source
We'll need a way to refresh the available studies from somewhere. The best place for now is remote settings.

[Some documentation that points the way for us to query that + avoid CORS](https://developer.chrome.com/extensions/xhr)

hamilton:
closes #2, closes #10


manifest.json
```diff
@@ -13,7 +13,11 @@
     }
   },
 
-  "permissions": ["telemetry"],
+  "permissions": [
+    "telemetry",
+    "storage",
+    "https://firefox.settings.services.mozilla.com/"
```
rhelmer, line 19: Should we have something in the `README` about this? There is a devserver that we can use for testing for instance: https://remote-settings.readthedocs.io/en/latest/tutorial-dev-server.html so it'd be nice to mention :)

rhelmer (changes requested):
I'm pretty new to svelte so I'm going to assume the way the storage is done is idiomatic etc, I don't see any problem with it. It looks a little odd to me to have `.then()` inside an `async` function with `fetch` for instance. I think we should at minimum console log any exceptions, if you know anyone who might be better at reviewing the svelte storage-specific stuff that'd be cool too.

rhelmer:
One thing we should think about is that we need to track all studies the user has been enrolled in, and why they ended (either we specified it on the server, or they ended the study themselves). This is so users can retroactively request their data to be deleted (issue #13).

I've been thinking that we would never remove a study from remote settings and always just append, and have the UI decide how it wants to display that... but happy to have this challenged if it doesn't make sense :)

src/create-store.js
```diff
@@ -0,0 +1,64 @@
...
+              draft.activeStudies = draft.activeStudies.filter(
+                (id) => id !== studyID
+              );
+            } else {
+              draft.activeStudies.push(studyID);
+            }
+          });
+        })
+        .catch((err) => {
+          // FIXME: do something here.
```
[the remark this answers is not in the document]
hamilton: That's fine for now. I think we should probably also think about what the UX should be for error-catching these types of requests.

hamilton:
re: those `.then` statements, I'm fine using basic async/await – I worked on a project recently where someone _didn't_ want it for whatever reason, so perhaps I've over-rotated a bit too much :)

manifest.json
hamilton: That would be nice. I'm not sure I fully follow – let's sync up on this so I understand the details

rhelmer (approved):

PR #12, opened 2020-10-05
Use the webextension-polyfill library instead of `chrome`

Dexterp37:
I acknowledge this is a bit hacky, especially in the `hooks.js` file.

### DIFF
--- a/core-addon/IonCore.js
+++ b/core-addon/IonCore.js
@@ -5,9 +5,9 @@
 module.exports = class IonCore {
   initialize() {
     // Whenever the addon icon is clicked, open the control page.
-    chrome.browserAction.onClicked.addListener(this._openControlPanel);
+    browser.browserAction.onClicked.addListener(this._openControlPanel);
     // After installing the addon, make sure to show the control page.
-    chrome.runtime.onInstalled.addListener(async ({ reason, temporary }) => {
+    browser.runtime.onInstalled.addListener(async ({ reason, temporary }) => {
       if (reason !== "install") {
         // We're only showing this when the addon is installed!
         return;
@@ -17,6 +17,6 @@ module.exports = class IonCore {
   }
 
   _openControlPanel() {
-    chrome.runtime.openOptionsPage();
+    browser.runtime.openOptionsPage();
   }
 }
--- a/core-addon/IonCore.js
+++ b/core-addon/IonCore.js
@@ -17,6 +17,8 @@ module.exports = class IonCore {
   }
 
   _openControlPanel() {
-    browser.runtime.openOptionsPage();
+    browser.runtime.openOptionsPage().catch(e => {
+      console.error(`IonCore.js - Unable to open the control panel`, e);
+    });
   }
 }
--- a/core-addon/background.js
+++ b/core-addon/background.js
@@ -2,7 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const IonCore = require('./IonCore.js');
+var browser = require("../node_modules/webextension-polyfill/dist/browser-polyfill.js");
+const IonCore = require("./IonCore.js");
 
 const ionCore = new IonCore();
 ionCore.initialize();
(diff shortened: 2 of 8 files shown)


core-addon/background.js
```diff
@@ -2,7 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-const IonCore = require('./IonCore.js');
+var browser = require("../node_modules/webextension-polyfill/dist/browser-polyfill.js");
```
Dexterp37, line 5: There must be some way to not use the full path here, but I failed to find it.

tests/core-addon/hooks.js
```diff
@@ -0,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var browser = require('sinon-chrome/extensions');
+// We need to provide the `browser.runtime.id` for sinon-chrome to
+// be happy and play nice with webextension-polyfill. See this issue:
+// https://github.com/mozilla/webextension-polyfill/issues/218
+browser.runtime.id = "testid";
+global.browser = browser;
```
Dexterp37, line 10: This was required, apparently the `beforeAll` hook was not sufficient.

core-addon/background.js
brizental: `const browser = require("webextension-polyfill");` doesn't work?

Dexterp37: This seems to work for the test part, but when packing the addon it fails with:

```
$ npm run build-addon

> svelte-app@1.0.0 build-addon /mnt/d/Mozilla/ion/ion-svelte
> rollup -c rollup.config.addon.js


core-addon/background.js → public/addon-build/background.js...
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
webextension-polyfill (imported by core-addon/background.js, webextension-polyfill?commonjs-external)
created public/addon-build/background.js in 167ms
```

tests/core-addon/hooks.js
brizental: In this case, there is no need for the `beforeAll` hook?

core-addon/background.js
Dexterp37: Thanks for the help! This is now fixed!

tests/core-addon/hooks.js
Dexterp37: It's still required. We want to flush/remove in `afterAll` and then reinstate in `beforeAll`. I suspect the first `global.browser` is used way earlier than running the tests!

core-addon/IonCore.js
```diff
@@ -17,6 +17,6 @@ module.exports = class IonCore {
   }
 
   _openControlPanel() {
-    chrome.runtime.openOptionsPage();
+    browser.runtime.openOptionsPage();
```
rhelmer, line 20: This returns a promise that might reject, do you want to either catch this (and log it I guess?) or maybe make `initialize` an async function?

rhelmer (changes requested):

core-addon/IonCore.js
rhelmer: How about this - could we store a boolean (`false` by default) somewhere (webextension storage probably) which is set to `true` if this page loads successfully? Then on startup of the extension, we can try to display the page again.

I can't think *why* exactly this would fail, but both MDN and the Chrome extension docs mention it: https://developer.chrome.com/extensions/runtime#method-openOptionsPage

Dexterp37: > How about this - could we store a boolean (`false` by default) somewhere (webextension storage probably) which is set to `true` if this page loads successfully? If it fails, then on startup of the extension, we can try to display the page again.

This seems like a very good idea. Do you mind if I take it to a separate, future PR while only catching it here?

rhelmer: wfm!

rhelmer: Filed issue #17

rhelmer (approved):

PR #18, opened 2020-10-07
Generate enrollment pings and add test coverage

### LINKED ISSUE #13: Send Ion enrollment pings
Joining Ion as well as individual studies should send an enrollment ping, and leaving studies or Ion should send deletion pings. The code in `about:ion` that does this is: https://searchfox.org/mozilla-central/source/browser/components/ion/content/ion.js#714-789

We should be able to do this with `browser.telemetry.submitEncryptedPing`, I think. We'll need to provide the key in the `manifest.json`: https://github.com/rhelmer/pioneer-v2-example/blob/63730cd/manifest.json#L18-L32

Dexterp37:
Fixes #13.

This came out to be a big bigger than I initially expected! Unfortunately the legacy telemetry APIs exposed to addons make a few assumptions (e.g. study name lives in the manifest). These assumptions do not hold true anymore as we move towards the new architecture. To work around this, I added an "experimental" API that exposes a custom legacy telemetry API to the addon.

See the individual commits for more information.

TODO (the deletion ping will be added as a separate PR):

- [x] Unit tests;
- [x] Study enrollment (we only have platform enrollment)

### DIFF
--- a/core-addon/IonCore.js
+++ b/core-addon/IonCore.js
@@ -2,6 +2,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+// The path of the embedded resource used to control Ion options.
+const ION_OPTIONS_PAGE_PATH = "public/index.html";
+
 module.exports = class IonCore {
   initialize() {
     // Whenever the addon icon is clicked, open the control page.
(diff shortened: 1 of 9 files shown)


core-addon/IonCore.js
```diff
@@ -14,11 +17,135 @@ module.exports = class IonCore {
...
+   * Enroll in the Ion platform.
+   *
+   * This sets up all the required information (e.g. Ion ID)
+   * and sets the relevant data to the pipeline.
+   *
+   * @returns {Promise} A promise resolved when the enrollment
+   *          is complete (does not block on data upload).
+   */
+  async _enroll() {
+    // TODO: Generate a proper random UUID.
```
Dexterp37, line 76: @rhelmer I ended up adding this as a safe placeholder to have the submission work. One thing we could do is just land this as it is, and follow up on proper UUID generation if you don't get to it :)

src/api/web-extension.js
```diff
@@ -58,10 +91,18 @@ export default {
...
-    return true;
+  /**
+   * Updates the overall Ion enrollment in the add-on.
+   * NOTE: if updating in the app store in the add-on suffices,
+   * this should probably just return the OK signal.
+   *
+   * @param {Boolean} enroll
+   *        `true` if user decided to enroll in the Ion platform,
+   *        `false` otherwise.
+   * TODO: Why can `enroll` be undefined?
```
Dexterp37, line 102: @hamilton hey! In `create-store.js` (`updateIonEnrollment`) we default `enroll = undefined`. When can that be undefined?

hamilton: I have no feeling about this. Please feel free to implement whatever you want in `web-extension.js` and have the return value be whatever format you prefer. We can then check that in `create-store.js` to make sure the frontend can respond appropriately. I have no guidance here other than feel free to own it.

core-addon/IonCore.js
rhelmer: I think it's OK to land as-is, I am going to attempt to use `browserify` so we could use https://github.com/uuidjs/uuid or something similar, that says it is compliant with the RFC. It's tempting to write our own but there are a few odd edge cases I'd like to avoid dealing with :)

core-addon/telemetry/LegacyTelemetryApi.js
```diff
@@ -0,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
rhelmer, line 1: We *could* alter the `browser.telemetry.submitEncryptedPing` API to accommodate this, but since we're moving to glean eventually I think this is a fine approach.

core-addon/IonCore.js
rhelmer: Until we get this figured out, you could add an experiments API that uses the Firefox UUID generator if you want? I took a look and it's implemented in C++ so probably not something we can pluck out easily ;)

src/api/web-extension.js
Dexterp37: Oh, I see. So it toggles if nothing is provided, otherwise it reflects what's in `enroll`! Thanks for clarifying, that's much clearer, I'll add some documentation so that I don't forget :)

core-addon/telemetry/LegacyTelemetryApi.js
Dexterp37: Yes, I was thinking the same: we will eventually get rid of that API or change it in Firefox next year (regardless of Ion). Moreover, changing the API this way has the nice advantage of not having to ride the trains :)

core-addon/IonCore.js
Dexterp37: > 
> 
> Until we get this figured out, you could add an experiments API that uses the Firefox UUID generator if you want? I took a look and it's implemented in C++ so probably not something we can pluck out easily ;)

Great idea! Let me do this!

src/api/web-extension.js
hamilton: Works for me!

tests/core-addon/IonCore.test.js
```diff
@@ -19,7 +20,138 @@ describe('IonCore', function () {
...
+      // ... an empty payload ...
+      assert.equal(Object.keys(submitArgs[1]).length, 0);
+      // ... and a specific set of options.
+      assert.equal(submitArgs[2].studyName, "pioneer-meta");
+      assert.equal(submitArgs[2].encryptionKeyId, "discarded");
+      assert.equal(submitArgs[2].schemaName, "pioneer-enrollment");
+      assert.equal(submitArgs[2].schemaNamespace, "pioneer-meta");
+    });
+
+    it('dispatchers study-enrollment messages', async function () {
```
hamilton, line 113:
```suggestion
    it('dispatches study-enrollment messages', async function () {
```

Dexterp37: GAH! Good catch :-) I'll address this right after reviews are complete, so that I can fix everything in one go!

hamilton: You can commit directly in this UI! Didn't want to make you do more work :)

core-addon/IonCore.js
```diff
@@ -14,11 +17,153 @@ module.exports = class IonCore {
...
+  /**
+   * Enroll in an Ion Study.
+   *
+   * This sends the required pings,
+   *
+   * @returns {Promise} A promise resolved when the enrollment
+   *          is complete (does not block on data upload).
+   */
+  async _enrollStudy(studyAddonId) {
+    // TODO: Validate the study id?
```
rhelmer, line 104: This is a good idea... maybe just that it's a valid UUID

rhelmer (approved):

core-addon/IonCore.js
Dexterp37: Yes! Are addon ids always uuids? I think I saw things such as "addon@mozilla.com"

Issue #3, opened 2020-10-01
configure web-ext to refresh if the Svelte bundle updates


hamilton:
From @rhelmer on slack:

> the one thing I feel like we're missing is the ability to have both svelte and the web extension reload seamlessly while running with web-ext... right now you have to npm run build then web-ext run, which we could have a single command for but you don't get the nice Svelte auto-reload

This can probably be surmounted easily.

rhelmer:
One way to do this is to have `rollup -w` running and then watch the build file output with `web-ext run --watch-file public/build/bundle.js`, this seems to work OK.

The way `web-ext` does reload is a little odd in that it's essentially an upgrade not a remove/reinstall of the add-on, so you need to e.g. reload the `moz-extension://` page to see changes if you already have it open, but that's normal `web-ext` behavior so not too surprising I don't think.

rhelmer:
> The way `web-ext` does reload is a little odd in that it's essentially an upgrade not a remove/reinstall of the add-on, so you need to e.g. reload the `moz-extension://` page to see changes if you already have it open, but that's normal `web-ext` behavior so not too surprising I don't think.

Looking into this a bit more, I'm not 100% sure how the "live reload" feature works but I suspect we could get that working as well, which would make the above a little nicer and more like the normal Svelte dev experience.

rhelmer:
> Looking into this a bit more, I'm not 100% sure how the "live reload" feature works but I suspect we could get that working as well, which would make the above a little nicer and more like the normal Svelte dev experience.

Hm, ok so I see that the svelte app inside the webextension is trying to load `<script async="" src="//5db84e64-6e70-184a-9cbd-14f433219489:35729/livereload.js?snipver=1" id="livereloadscript"></script>` but the file isn't present in the extension, I suspect that if we fix this it might Just Work ™

hamilton:
fwiw I think rollup just generates the following w/ communicates w/ the dev server:

```javascript
(function(l, r) { if (l.getElementById('livereloadscript')) return; r = l.createElement('script'); r.async = 1; r.src = '//' + (window.location.host || 'localhost').split(':')[0] + ':35729/livereload.js?snipver=1'; r.id = 'livereloadscript'; l.getElementsByTagName('head')[0].appendChild(r) })(window.document);
```

hamilton:
Another option would be to skip the `livereload` from rollup and use [the actual LiveReload](https://github.com/napcs/node-livereload) in some way.

[...]

PR #28, opened 2020-10-09
document building from a clean repo to a generated add-on artifact

rhelmer:
I noticed that the current instructions don't work on a clean checkout (`npm run build-addon` must be run before `web-ext`). @hamilton and I discussed and the real fix here is to simplify these build targets, but for the moment I think having them work out-of-the-box is good enough.

I've also added instructions on how to generate an add-on file and what to do with it.

These are the instructions I used to generate the 0.1 release .xpi

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,14 +23,15 @@ To run a local webserver (`http://localhost:5000`) for the UI which auto-reloads
 npm run dev
 ```
 
+# Build the Core addon
+
 When ready, package up for use inside the web extension:
 
 ```bash
 npm run build
 ```
 
-# Build the Core addon
-With the UI is generated, the addon can finally be built:
+Build the local version of the addon:
 
 ```bash
 npm run build-addon
--- a/README.md
+++ b/README.md
@@ -29,6 +29,13 @@ When ready, package up for use inside the web extension:
 npm run build
 ```
 
+# Build the Core addon
+With the UI is generated, the addon can finally be built:
+
+```bash
+npm run build-addon
+```
+
 ...then start [web-ext](https://github.com/mozilla/web-ext):
 
 ```bash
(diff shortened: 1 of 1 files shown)


rhelmer:
This is the 0.1 release: https://github.com/mozilla-ion/ion-core-addon/releases/tag/0.1

We're doing this now so we can get a website up on github.io and have something to point to, as a start for allowing other folks to dogfood. We're going to hit issues with signatures and having to use non-release builds and so on but that's fine, it's a start :)

Dexterp37 (approved):

Dexterp37:
> 
> 
> I noticed that the current instructions don't work on a clean checkout (`npm run build-addon` must be run before `web-ext`). @hamilton and I discussed and the real fix here is to simplify these build targets, but for the moment I think having them work out-of-the-box is good enough.

Oh, good point. We should probably a single build target. I believe that, to simplify addon development life, we should also have a target to exclusively build the addon, since building the UI takes a few seconds more (and when you're quickly iterating this is annoying!)

rhelmer:
> > I noticed that the current instructions don't work on a clean checkout (`npm run build-addon` must be run before `web-ext`). @hamilton and I discussed and the real fix here is to simplify these build targets, but for the moment I think having them work out-of-the-box is good enough.
> 
> Oh, good point. We should probably a single build target. I believe that, to simplify addon development life, we should also have a target to exclusively build the addon, since building the UI takes a few seconds more (and when you're quickly iterating this is annoying!)

For dev purposes we have been investigating issue #3, and thinking we'd have two high-level build targets (one for hot-reloading dev, and one for generating the final extension artifact).

We could have these composed of smaller targets too but I suspect the hot-reloading will be fast enough we won't care :)

PR #29, opened 2020-10-09
adds demo site

### LINKED ISSUE #8: add scaffolding for the Ion site
We've agreed to put together the site implementation in this repository. This enables us to (1) do end-to-end testing cohesively, and (2) share UX components that are in both the add-on and the site itself.

For now, let's just use `docs/` and add a static html page until we have the full UX put together.

hamilton:
Closes #8 ~ This is just meant to be enough to demonstrate an end-to-end test.

Screen Shot 2020-10-08 at 9 44 58 PM


docs/common.css
```diff
@@ -0,0 +1,760 @@
...
+.back-button:dir(rtl) {
+  transform: scaleX(-1);
+}
+
+xul|tabs {
+  padding-inline: 0;
+  position: static;
+}
+
+xul|tab[visuallyselected] {
```
rhelmer, line 682: Oh is this the CSS that `about:ion` uses? I was wondering why I was seeing `xul` everywhere, I think this is OK for now but let's ditch this CSS file soon.

docs/index.html
```diff
@@ -0,0 +1,186 @@
...
+      Put your data to work for a better internet.
+    </h1>
+    <a
+      tabindex="0"
+      href="https://github.com/mozilla-ion/ion-core-addon/releases/download/0.1/ion_demo_study-2.0.xpi"
+      >Install the Ion Web Extension</a
+    >
+    <p>
+      Note: for this demoware, you're going to need to run Nightly & set
+      <code>xpinstall.signatures.required = false</code> in
```
rhelmer, line 182: This is great!

rhelmer (approved):
Looks awesome let's do it

docs/index.html
rhelmer: Oh hm actually there's another pref you'll want here... `extensions.experiments.enabled = true`

rhelmer: Also there will be a third (ugh) once we get the signing infra stuff nailed down, `xpinstall.signatures.dev-root = true` (this one does not have a default and must be created)

rhelmer:
So I think this is fine to merge now, with the expectation that the add-on will load but the privileged part won't function until we figure out how best to use the signing infra. I think ultimately we'll need three prefs to QA this mentioned in the review.

docs/common.css
hamilton: sorry! Yes, it is, but honestly I might get rid of this since I just went all-in

docs/index.html
hamilton: added. Thanks!

Issue #27, opened 2020-10-08
npm commands should just be dev & build for now


hamilton:
Our `readme` is pretty complicated. Ideally someone can just run `npm run dev` or `npm run build` and get the expected behavior.

Another possibility: `npm run smoke-test` or something.

Dexterp37:
Morphed this issue to be about consolidating all the npm commands: we have test-addon, build, build-addon, lint-addon, lint. We should spec out and simplify them all.

PR #30, opened 2020-10-09
Send the deletion pings for all the studies user was enrolled in

### LINKED ISSUE #23: Send Ion deletion pings
Other half of #13.

Dexterp37:
Fixes #23

This PR changes a few things::

- it refactors the code to send empty pings, to share it between enrollment and deletion;
- it abstracts away the code to interact with the local storage;

### DIFF
--- a/core-addon/IonCore.js
+++ b/core-addon/IonCore.js
@@ -66,6 +66,9 @@ module.exports = class IonCore {
         // is expecting it.
         return this._enrollStudy(message.data.studyID).then(r => true);
       } break;
+      case "unenrollment": {
+        return this._unenroll().then(r => true);
+      } break;
       default:
         return Promise.reject(
           new Error(`IonCore - unexpected message type ${message.type}`));
@@ -114,6 +117,36 @@ module.exports = class IonCore {
 
     // Finally send the ping.
     await this._sendEnrollmentPing(studyAddonId);
+  }
+
+  /**
+   * Unenroll from the Ion platform.
+   *
+   * This clears all the stored data (e.g. Ion ID)
+   * and sends the relevant deletion requests to the pipeline.
+   *
+   * @returns {Promise} A promise resolved when the unenrollment
+   *          is complete (does not block on data upload).
+   */
+  async _unenroll() {
+    // Read the list of the studies user activated throughout
+    // their stay on the Ion platform and send a deletion request
+    // for each of them.
+    let studyList = await this._storage.getActivatedStudies();
+    for (let studyId of studyList) {
+      await this._sendDeletionPing(studyId);
+    }
+
+    // Clear locally stored Ion ID.
+    await this._storage.clearIonID();
+
+    // The telemetry API, before sending a ping, reads the
+    // ion id from a pref. We're good to clear this after sending
+    // the deletion pings.
+    await browser.legacyTelemetryApi.clearIonID();
+
+    // Finally clear the list of studies user took part in.
+    await this._storage.clearActivatedStudies();
   }
 
   /**
--- a/core-addon/IonCore.js
+++ b/core-addon/IonCore.js
@@ -108,22 +108,27 @@ module.exports = class IonCore {
   }
 
   /**
-   * Sends a Pioneer enrollment ping.
+   * Sends an empty Ion ping with the provided info.
    *
-   * The `creationDate` provided by the telemetry APIs will be used as the timestamp
-   * for considering the user enrolled in pioneer and/or the study.
+   * @param {String} payloadType
+   *        The type of the encrypted payload. This will define the
+   *        `schemaName` of the ping.
    *
-   * @param [studyAddonid=undefined] - optional study id. It's sent in the ping, if
-   *        present, to signal that user enroled in the study.
+   * @param {String} namespace
+   *        The namespace to route the ping. This will define the
+   *        `schemaNamespace` and `studyName` properties of the ping.
    */
-  async _sendEnrollmentPing(studyAddonId) {
+  async _sendEmptyPing(payloadType, namespace) {
     let options = {
-      studyName: "pioneer-meta",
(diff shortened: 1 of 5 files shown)


rhelmer:
Now that we're using privileged APIs here I think we should modify the README slightly to specify: `web-ext run --pref=extensions.experiments.enabled=true -f nightly`

We can hide this behind better build targets in issue #27 but for now that makes this work out-of-the-box without a lot of `about:config` fiddling and so on.

core-addon/IonCore.js
```diff
@@ -166,4 +204,42 @@ module.exports = class IonCore {
...
+  }
+
+  /**
+   * Sends a Ion deletion-request ping.
+   *
+   * @param {String} studyAddonid
+   *        It's sent in the ping to signal that user unenroled from a study.
+   */
+  async _sendDeletionPing(studyAddonId) {
+    if (typeof studyAddonId == "undefined") {
```
rhelmer, line 239: I think [strict equality operator for checking undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined) is better here since you don't need to use a string (and have a potential typo etc). `(studyAddonId === undefined)` will throw if you misspell "undefined" in other words, while `typeof studyAddonId == "undefnied"` will just do the wrong thing :)

core-addon/IonCore.js
```diff
@@ -166,4 +204,42 @@ module.exports = class IonCore {
...
+    // If this is an enrollment ping for the pioneer program (in contrast to the enrollment to
+    // a specific study), use a meta namespace.
+    return await this._sendEmptyPing("pioneer-enrollment", "pioneer-meta");
+  }
+
+  /**
+   * Sends a Ion deletion-request ping.
+   *
+   * @param {String} studyAddonid
+   *        It's sent in the ping to signal that user unenroled from a study.
```
rhelmer, line 236: `s/unenroled/unenrolled/`

core-addon/IonCore.js
```diff
@@ -60,6 +66,9 @@ module.exports = class IonCore {
         // is expecting it.
         return this._enrollStudy(message.data.studyID).then(r => true);
       } break;
+      case "unenrollment": {
+        return this._unenroll().then(r => true);
```
rhelmer, line 70: I'm a little confused by this idiom... is this saying that if `_unenroll()` doesn't throw anything then return `true` essentially? I think it's OK but seems like we really want `initialize()` to be an `async` function or something.

Since there's no `catch` then this means the exception is basically unhandled?

core-addon/Storage.js
```diff
@@ -0,0 +1,65 @@
...
+  /**
+   * Adds a study id to the stored list of activated studies.
+   *
+   * @param {String} studyId
+   *        The id of the study to add to the list. If the id
+   *        is already present, this function is a no-op.
+   * @returns {Promise} resolved with the list of activated studies
+   *          if the study is added to the list, rejected on errors.
+   */
+  async appendActivatedStudy(studyId) {
```
rhelmer, line 45: We don't need to do it here necessarily but one thing we should keep track of is if a study was ended either by the user or the server, because they may not be re-joinable. This is complicated enough I will just file another issue for it.

rhelmer (approved):
I have a few questions and suggestions but it looks great to me overall!

core-addon/IonCore.js
Dexterp37: > 
> 
> I'm a little confused by this idiom... is this saying that if `_unenroll()` doesn't throw anything then return `true` essentially?

I'm saying "only return true if `_unenroll` is successful". On problems, the exception bubbles up to the UX side.

>I think it's OK but seems like we really want `initialize()` to be an `async` function or something.

Why do we need `initialize()` to be `async`? Did you mean `_handleMessage`? This function is basically always returning a promise (either resolved or rejected). The result of such promise is returned to the sender. This is the [recommended way advertised on our MDN docs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage).

> Since there's no `catch` then this means the exception is basically unhandled?

Good catch. This is basically saying "it's a UX problem". But you nailed the underlying bigger problem: how are we handling errors or things that go wrong on the UI? The current approach makes the UI aware, but this means that the UI will break if no error is handled.

Should we file an issue and deal about this problem separately? It seems like a big topic on its own.

core-addon/IonCore.js
```diff
@@ -103,52 +112,81 @@ module.exports = class IonCore {
   async _enrollStudy(studyAddonId) {
     // TODO: Validate the study id?
```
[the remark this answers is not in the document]
Dexterp37: Oh, definitely! I'm planning on doing this as part of #33 !

core-addon/IonCore.js
Dexterp37: /me fails. Great catch :)

core-addon/IonCore.js
Dexterp37: OH! Got it! That's a great tip, thanks for the explanation!

src/api/web-extension.js
```diff
@@ -17,6 +17,7 @@ async function sendToCore(type, payload) {
   const VALID_TYPES = [
     "enrollment",
     "study-enrollment",
+    "unenrollment",
```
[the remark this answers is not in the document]
Dexterp37: Right now, we don't do anything on `study-unenrollment`. This is something we'll need to do: the deletion pings are only sent when unenrolling from the whole Ion platform.

core-addon/Storage.js
Dexterp37: Thanks!

core-addon/IonCore.js
Dexterp37: Created #37!

PR #35, opened 2020-10-09
web-ext command to enable experiments

rhelmer:
Should remove the need to do additional configuration in `about:config` etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,10 +40,10 @@ npm run build-addon
 ...then start [web-ext](https://github.com/mozilla/web-ext):
 
 ```bash
-web-ext run
+web-ext run --pref=extensions.experiments.enabled=true -f nightly
 ```
 
-`web-ext` defaults to Firefox, it can be run with Chrome:
+`web-ext` uses Nightly Firefox above, it can be run with Chrome:
 
 ```bash
 web-ext run -t chromium


hamilton:
@rhelmer should this be an `npm` command? Might make it maximally easy.

rhelmer:
> @rhelmer should this be an `npm` command? Might make it maximally easy.

@hamilton Hm this is a good idea... I think we shouldn't necessarily advertise cross-browser compatibility yet since we have some privileged features that are Firefox-only. So in the future we could support `npm run dev -t chromium` or something, but have `npm run dev` use Firefox Nightly (this is the default behavior of `web-ext`)?

I was thinking through what it might look like (before I write any code ;)), what do you think about:

## Get started

Install the dependencies:

```bash
cd ion-core-addon
npm install
```

## Run the extension in a browser in dev mode

Run the extension in Firefox Nightly:

```bash
npm run dev
```
To run test coverage:

```bash
npm run test
```

## Produce a production WebExtension

```bash
npm run build
```

The output will be a `.zip` file in `web-ext-artifacts/`, this can be renamed to `.xpi` for Firefox or `.crx` for Chrome.

hamilton:
@rhelmer that looks good to me. We can revise as needed in the future.

Dexterp37:
@rhelmer is this still relevant?

rhelmer:
This change was already made to the README and we can handle cleaning up build targets in issue #27

PR #38, opened 2020-10-12
Rename the privileged API object and fix study leave button

Dexterp37:
See the individual commits for the explanation.


hamilton (approved):
This works as expected.

PR #51, opened 2020-10-14
Fix the dependencies errors on CircleCI

Dexterp37:


Dexterp37:
All right, turns out I will need to merge this too as it is for things to properly show up in the branch protection options.

PR #62, opened 2020-10-14
Send an "uninstall" message when leaving a study

### LINKED ISSUE #58: when a user clicks "leave study", the core add-on should send a message to the study to uninstall

Dexterp37:
Fixes #58 

This allows studies to be uninstalled when clicking on the "Leave" button from the Ion Core addon.

### DIFF
--- a/core-addon/IonCore.js
+++ b/core-addon/IonCore.js
@@ -169,8 +169,15 @@ module.exports = class IonCore {
       return Promise.reject(
         new Error(`IonCore._unenrollStudy - Unknown study ${studyAddonId}`));
     }
-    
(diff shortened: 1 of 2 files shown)


hamilton (approved):
The code is minimal, so easy to review. I ended up rebasing this on my branch that allows us to define our own available studies, which I used in conjunction with manually loading the basic study add-on, and then tested the uninstall ping. I have never signed an add-on so I didn't go through generating an xpi, probably easy but I will have to revisit tomorrow. A lot of jumping through hoops but it works as demonstrated in the below gif. Notice the two add-ons. I click "leave study", and lo and behold, only one add-on left in `about:addons`.

ion-uninstall-study-02

PR #64, opened 2020-10-14
introduces a new local option for the addon build process that allows for locally-available studies list in /public

### LINKED ISSUE #44: Make the URL from which we fetch studies build-time configurable
We currently store the URL we query to fetch the available studies both in the `manifest.json` and in the code used to perform the actual `fetch`. We might want this URL to be customizable for QA purposes and one potential solution to this could be to make this URL a Rollup build time configuration option and expose it.

hamilton:
Closes #44 

This _in  theory_ makes it so we can easily build the add-on to point toward a local "available studies" json payload instead of the Remote Settings list. This allows us ways to test study add-ons locally or point to other xpis. It does this by:

- leveraging `@rollup/plugin-replace` to change `core-addon/background.js` to have a swappable URL
- makes `npm run build-addon` use the Remote Settings URI via the above
- makes `npm run local-build-addon` use `/public/locally-available-studies.json` as the target, which can be used by the user to test out any locally-built add-ons.

This isn't probably the most optimal solution in the long run, but I think it should work well for now & allow us to showcase an end-to-end experience w/ the Basic Add-On.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,3 +62,15 @@ web-ext build
 ```
 
 The output will be a `.zip` file in `web-ext-artifacts/` - this should be renamed to `.xpi` for Firefox and `.crx` for Chrome.
+
+## Building and testing a study locally
+
+To run an end-to-end local test with your own study add-on, first build your study (if you don't have one, you can [build the Ion Basic Study](https://github.com/mozilla-ion/ion-basic-study)) and export the build as `<name-of-study>.xpi`. Edit `/public/locally-available-studies.json` so that `sourceURI.spec` is `/public/<name-of-study>.xpi` (you can change the other fields in `/public/locally-available-studies.json` as well for demo purposes as needed). 
+
+Then run:
+
+- `npm run build`
+- `npm run demo-build-addon`
+- `web-ext run --pref=extensions.experiments.enabled=true -f nightly`
+
+To walk through the Core Add-On experience with your study.
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ The output will be a `.zip` file in `web-ext-artifacts/` - this should be rename
 
 ## Building and testing a study locally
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -61,4 +61,16 @@ To generate an installable extension file:
 web-ext build
 ```
 
-The output will be a `.zip` file in `web-ext-artifacts/` - this should be renamed to `.xpi` for Firefox and `.crx` for Chrome.
\ No newline at end of file
+The output will be a `.zip` file in `web-ext-artifacts/` - this should be renamed to `.xpi` for Firefox and `.crx` for Chrome.
+
+## Building and testing a study locally
+
+To run an end-to-end local test with your own study add-on, first build your study (if you don't have one, you can [build the Ion Basic Study](https://github.com/mozilla-ion/ion-basic-study)) and export the build as `<name-of-study>.xpi`. Edit `/public/locally-available-studies.json` so that `sourceURI.spec` is `/public/<name-of-study>.xpi` (you can change the other fields in `/public/locally-available-studies.json` as well for demo purposes as needed).
```
hamilton, line 68: heh, I just discovered that you cannot in fact do this very easily :)

core-addon/IonCore.js
```diff
@@ -7,8 +7,14 @@ const Storage = require("./Storage.js");
 // The path of the embedded resource used to control Ion options.
 const ION_OPTIONS_PAGE_PATH = "public/index.html";
 
+const ION_DEFAULT_ARGS = {
+  availableStudiesURI: "https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/pioneer-study-addons-v1/records"
+}
+
 module.exports = class IonCore {
-  constructor() {
+  constructor(args = {}) {
```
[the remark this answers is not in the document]
hamilton: will do

README.md
hamilton: The part that doesn't "work" is making an xpi. I just haven't doen it before, so my instructions were not quite accurate.

core-addon/IonCore.js
```diff
@@ -7,8 +7,19 @@ const Storage = require("./Storage.js");
 // The path of the embedded resource used to control Ion options.
 const ION_OPTIONS_PAGE_PATH = "public/index.html";
 
+const ION_DEFAULT_ARGS = {
+  availableStudiesURI: "https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/pioneer-study-addons-v1/records"
```
Dexterp37, line 11: Please add a comment stating that if this URL ever changes in production, the `manifest.json` file would need to be changed as well.

Dexterp37 (approved):

PR #67, opened 2020-10-15
Allow Ion studies to send telemetry through the Core Addon

### LINKED ISSUE #47: implement message-passing from study to core for Telemetry
We need to pass [telemetry from the unprivileged add-on](https://github.com/mozilla-ion/ion-basic-study/pull/4#discussion_r504109099) to the Ion Core add-on.

We'll also need to figure out how to pass [the public key](https://github.com/mozilla-ion/ion-basic-study/pull/4#discussion_r504109549) from the study add-on, ideally not along with every single message :)

Dexterp37:
Fixes #47 .

This PR adds the ability to receive messages from known Ion Studies and does a bit of refactoring to move data collection code to its own module.

It also adds a message to send telemetry from an Ion study.

**DO NOT MERGE**: It's missing test coverage for the "telemetry-ping" message. I'll add that tomorrow morning.

### DIFF
--- a/core-addon/DataCollection.js
+++ b/core-addon/DataCollection.js
@@ -15,48 +15,28 @@ module.exports = class DataCollection {
    *        `schemaNamespace` and `studyName` properties of the ping.
    */
   async _sendEmptyPing(payloadType, namespace) {
-    let options = {
-      studyName: namespace,
-      addPioneerId: true,
-      // NOTE - while we're not actually sending useful data in
-      // this payload, the current Ion v2 Telemetry pipeline requires
-      // that pings are shaped this way so they are routed to the correct
-      // environment.
-      //
-      // At the moment, the public key used here isn't important but we do
-      // need to use *something*.
-      encryptionKeyId: "discarded",
-      publicKey: {
-        crv: "P-256",
-        kty: "EC",
-        x: "XLkI3NaY3-AF2nRMspC63BT1u0Y3moXYSfss7VuQ0mk",
-        y: "SB0KnIW-pqk85OIEYZenoNkEyOOp5GeWQhS1KeRtEUE",
-      },
-      schemaName: payloadType,
-      schemaVersion: 1,
-      // Note that the schema namespace directly informs how data is
-      // segregated after ingestion.
-      // If this is an enrollment ping for the pioneer program (in contrast
-      // to the enrollment to a specific study), use a meta namespace.
-      schemaNamespace: namespace,
+    // NOTE - while we're not actually sending useful data in
+    // this payload, the current Ion v2 Telemetry pipeline requires
+    // that pings are shaped this way so they are routed to the correct
+    // environment.
+    //
+    // At the moment, the public key used here isn't important but we do
+    // need to use *something*.
+    const publicKey = {
+      crv: "P-256",
+      kty: "EC",
+      x: "XLkI3NaY3-AF2nRMspC63BT1u0Y3moXYSfss7VuQ0mk",
+      y: "SB0KnIW-pqk85OIEYZenoNkEyOOp5GeWQhS1KeRtEUE",
     };
 
-    // For enrollment, we expect to send an empty payload.
-    const payload = {};
-
-    // We intentionally don't wait on the promise returned by
-    // `submitExternalPing`, because that's an internal API only meant
-    // for telemetry tests. Moreover, in order to send a custom schema
-    // name and a custom namespace, we need to ship a custom "experimental"
-    // telemetry API for legacy telemetry.
-    await browser.firefoxPrivilegedApi
-      .submitEncryptedPing("pioneer-study", payload, options)
-      .then(() => {
-        console.debug(`IonCore._sendEnrollmentPing - options: ${JSON.stringify(options)} payload: ${JSON.stringify(payload)}`);
-      })
-      .catch(error => {
-        console.error(`IonCore._sendEnrollmentPing failed - error: ${error}`);
-      });
+    await this.sendPing(
+      payloadType,
+      // We expect to send an empty payload.
+      {},
+      namespace,
+      "discarded",
+      publicKey
+    );
   }
 
   /**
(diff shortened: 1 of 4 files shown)


core-addon/DataCollection.js
```diff
@@ -0,0 +1,128 @@
...
+   */
+  async _sendEmptyPing(payloadType, namespace) {
+    // NOTE - while we're not actually sending useful data in
+    // this payload, the current Ion v2 Telemetry pipeline requires
+    // that pings are shaped this way so they are routed to the correct
+    // environment.
+    //
+    // At the moment, the public key used here isn't important but we do
+    // need to use *something*.
+    const publicKey = {
```
rhelmer, line 25: I wonder if this key should be configurable? It's also still kinda odd to me that we need to do this :)

core-addon/DataCollection.js
```diff
@@ -0,0 +1,128 @@
...
+   *
+   * @param {String} [studyAddonid=undefined]
+   *        optional study id. It's sent in the ping, if present, to signal
+   *        that user enroled in the study.
+   */
+  async sendEnrollmentPing(studyAddonId) {
+    // If we were provided with a study id, then this is an enrollment to a study.
+    // Send the id alongside with the data and change the schema namespace to simplify
+    // the work on the ingestion pipeline.
+    if (typeof studyAddonId != "undefined") {
```
rhelmer, line 56:
```suggestion
    if (studyAddonId !== undefined) {
```

core-addon/DataCollection.js
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  /**
+   * Sends a Ion deletion-request ping.
+   *
+   * @param {String} studyAddonid
+   *        It's sent in the ping to signal that user unenrolled from a study.
+   */
+  async sendDeletionPing(studyAddonId) {
+    if (typeof studyAddonId === undefined) {
```
rhelmer, line 73: This won't do what you want because `typeof studyAddonId` will be `"undefined"` not `undefined` :)
```suggestion
    if (studyAddonId === undefined) {
```

core-addon/DataCollection.js
```diff
@@ -0,0 +1,128 @@
...
+      // Note that the schema namespace directly informs how data is
+      // segregated after ingestion.
+      schemaNamespace: namespace,
+    };
+
+    // We intentionally don't wait on the promise returned by
+    // `submitExternalPing`, because that's an internal API only meant
+    // for telemetry tests. Moreover, in order to send a custom schema
+    // name and a custom namespace, we need to ship a custom "experimental"
+    // telemetry API for legacy telemetry.
```
rhelmer, line 118: We *could* change the `browser.telemetry.submitEncryptedPing` API so that it behaves the way we want here. I think whether that is worth it depends on how long we expect glean.js to take :)

rhelmer: The fewer custom APIs we ship in the add-on the better, since there's actual test coverage in-tree for `browser.telemetry` so we won't be caught by surprise if there is some refactoring to the built-in telemetry in Firefox.

core-addon/IonCore.js
```diff
@@ -47,6 +49,25 @@ module.exports = class IonCore {
     };
     browser.management.onInstalled.addListener(addonStateHandler);
     browser.management.onUninstalled.addListener(addonStateHandler);
+
+    // Listen for incoming messages from the studies.
+    //
+    // ***IMPORTANT***:
```
rhelmer, line 55: Thank you for commenting this thoroughly!

core-addon/IonCore.js
```diff
@@ -103,6 +124,45 @@ module.exports = class IonCore {
...
+   *
+   * IMPORTANT: This is `async` and will always handle all the
+   * external messages coming from the studies. It will not allow
+   * any other registered listener to catch messages. See
+   * the notes at the listener registration site.
+   *
+   * @param {Object} message
+   *        The payload of the message.
+   * @param {runtime.MessageSender} sender
+   *        An object containing informations about who sent
```
rhelmer, line 138:
```suggestion
   *        An object containing information about who sent
```

core-addon/IonCore.js
```diff
@@ -103,6 +124,45 @@ module.exports = class IonCore {
...
+      return Promise.reject(
+        new Error(`IonCore._handleExternalMessage - unexpected sender ${sender.id}`));
+    }
+
+    switch (message.type) {
+      case "telemetry-ping": {
+        const data = message.data;
+        return await this._dataCollection.sendPing(
+          data.payloadType, data.payload, data.namespace,
+          data.keyId, data.key
```
rhelmer, line 157: This is a nit feel free to ignore :) 

```suggestion
        const {payloadType, payload, namespace, keyId, key} = message.data;
        return await this._dataCollection.sendPing(
          payloadType, payload, namespace,
          keyId, key
```

rhelmer: You could also just destructure inside the function call but I think that is probably harder to read:
```
        return await this._dataCollection.sendPing(
          {payloadType, payload, namespace,
          keyId, key} = message.data
```

rhelmer (approved):
I think the only thing that must be changed is the `undefined` check in `sendDeletionPing`, feel free to merge as long as you make this change (or alternatively, show me why I'm wrong!)

core-addon/DataCollection.js
Dexterp37: I'd rather hide this, it's really an implementation detail/something we do to make Pipeline happy. But we can revisit this if you feel strongly about it!

Yes, I agree that this is _really odd_. But probably not worth fixing before moving to GleanJS :)

core-addon/DataCollection.js
Dexterp37: Gah! Sorry about that, you pointed it out already in another PR.

core-addon/DataCollection.js
Dexterp37: This is a good point! Note, though, that `submitEncryptedPing` is basically sugar on the top of `TelemetryController.submitExternalPing`, which is a stable API surface. I would hardly believe that's going to change anytime soon! But, even if it does, the telemetry team (of which I'm part ;-) ) usually sends a PSA email long before that happens.

I'd recommend sticking to the current plan and tackling this if we have urgent needs as this requires trains-work!

core-addon/IonCore.js
Dexterp37: Oh, my pleasure and duty! I spent some time figuring it out and asking Addons folks, so it was the least I could do!

core-addon/IonCore.js
Dexterp37: Oh! That looks great! Let me do what you recommended in the first comment!

Dexterp37:
> 
> 
> I think the only thing that must be changed is the `undefined` check in `sendDeletionPing`, feel free to merge as long as you make this change (or alternatively, show me why I'm wrong!)

You're not wrong of course, I was :-) Made all the changes you requested (including nits!) and added some more tests and docs.

PR #69, opened 2020-10-15
point default local remotesettings file at ion basic add-on on AMO

rhelmer:
@hamilton note that this also changes the XPI URL to point to AMO, if you want to keep it inside the `ion-core-addon` repo that's OK too.

### DIFF
--- a/public/locally-available-studies.json
+++ b/public/locally-available-studies.json
@@ -12,14 +12,14 @@
         "url": "https://addons.mozilla.org/en-US/firefox/addon/ion-basic-study/",
         "name": "Ion Developers"
       },
-      "version": "2.0",
+      "version": "2.1",
       "addon_id": "ion-basic-study@mozilla.org",
       "moreInfo": {
         "spec": "https://addons.mozilla.org/en-US/firefox/addon/ion-basic-study/"
       },
       "isDefault": false,
       "sourceURI": {
-        "spec": "https://addons.mozilla.org/firefox/downloads/file/3659906/ion_basic_study-2.0-fx.xpi"
+        "spec": "https://addons.mozilla.org/firefox/downloads/file/3661128/ion_basic_study-2.1-fx.xpi"
       },
       "studyType": "extension",
       "studyEnded": false,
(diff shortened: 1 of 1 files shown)


public/locally-available-studies.json
```diff
@@ -3,29 +3,29 @@
     {
       "name": "Ion Base Study",
       "icons": {
-        "32": "https://addons.cdn.mozilla.net/user-media/addon_icons/2644/2644632-32.png?modified=4a64e2bc",
-        "64": "https://addons.cdn.mozilla.net/user-media/addon_icons/2644/2644632-64.png?modified=4a64e2bc",
-        "128": "https://addons.cdn.mozilla.net/user-media/addon_icons/2644/2644632-128.png?modified=4a64e2bc"
+        "32": "https://addons.cdn.mozilla.net/static/img/addon-icons/default-32.png",
```
rhelmer, line 6: We are currently using the default icon, this will need to change once we update that.

public/locally-available-studies.json
```diff
@@ -3,29 +3,29 @@
...
+      "version": "2.0",
+      "addon_id": "ion-basic-study@mozilla.org",
       "moreInfo": {
-        "spec": "https://addons.mozilla.org/en-US/firefox/addon/pioneer-v2-example/"
+        "spec": "https://addons.mozilla.org/en-US/firefox/addon/ion-basic-study/"
       },
       "isDefault": false,
       "sourceURI": {
-        "spec": "/public/ion-base-addon.xpi"
+        "spec": "https://addons.mozilla.org/firefox/downloads/file/3659906/ion_basic_study-2.0-fx.xpi"
```
rhelmer, line 22: This and the version number need to change whenever we update the study add-on on AMO. One benefit here though is that we can point it at pre-release versions and test that way before making add-on updates live.

rhelmer:
Hm also before we merge this (assuming you are OK point to AMO) I'd like to push an update to it, since we'll need to update the XPI URL.

hamilton (approved):

PR #78, opened 2020-10-16
Ensure `this` exists when receiving external messages

Dexterp37:
This fixes the `_handleExternalMessage` handler registration, which was mistakenly not binding the `this` context. This unblocks sending telemetry from the basic study.


rhelmer (approved):
Makes sense to me - had to re-read the comment to understand why we are using an `async` handler here so that's ok :) Arrow function should handle the `this` binding the way you want I think.

PR #94, opened 2020-10-19
adds storybook + one example to verify that it works as expected

### LINKED ISSUE #88: set up storybook to work through UX component implementations

hamilton:
Closes #88 

A good way to review this is to run `npm install` then `npm run storybook`. It should "work" – localhost:6060 should show a single story about the StudyCard.

Screen Shot 2020-10-19 at 11 10 52 AM


stories/StudyCardView01.svelte
```diff
@@ -0,0 +1,81 @@
+<script>
+  import StudyCard from "../src/StudyCard.svelte";
+
+  const studies = [
+    {
+      case: "Not enrolled in either ion nor study",
```
rhelmer, line 6: Hm pairing "either" and "nor" breaks my brain :) This maybe:
```suggestion
      case: "Enrolled in neither Ion nor study",
```

rhelmer (approved):

stories/StudyCardView01.svelte
rhelmer: "Enrolled in neither" isn't super fun to look at either... I think "Not enrolled in Ion or study" would be simpler and clearer still.

hamilton, line 6:
```suggestion
      case: "Not enrolled in Ion nor study",
```

PR #95, opened 2020-10-19
adds @mozilla-protocol/protocol

### LINKED ISSUE #6: integrate the Protocol Design System via npm
Ryan gave a heads-up that his UX design will likely require us to integrate [Protocol](https://protocol.mozilla.org/) into the web extension. Even after we get creative mocks for the web, we're still likely to leverage Protocol. Requirements on our end:

- [x] set up scss to support how Protocol works
- [x] pull in our current fonts (zilla + inter)
- [ ] have a layout component created to govern how the pages work
- [ ] remove existing CSS that clashes (`global.css` and `common.css`)
- [x] put together storybook for rapid iteration (#88)

hamilton:
closes #6 

This does the basic setup to support using the Protocol design system.~ Built on #94 – once that merges I'll rebase this on `master`.~ This PR is ready for review.

Screen Shot 2020-10-19 at 1 39 52 PM

### DIFF
--- a/src/components/Button.svelte
+++ b/src/components/Button.svelte
@@ -1,0 +1,31 @@
+<script>
+  import { get_current_component as getComponent } from "svelte/internal";
+  import { forwardEvents } from "./forwardEvents";
+  export let size = "md";
+  export let neutral = false;
+  export let product = false;
+  export let secondary = false;
+  export let dark = false;
+
+  export let disabled = false;
+
+  const forwardAll = forwardEvents(getComponent());
+
+  $: sizeClass = size ? `mzp-t-${size}` : undefined;
+  $: productClass = product ? `mzp-t-product` : undefined;
+  $: neutralClass = neutral ? `mzp-t-neutral` : undefined;
+  $: secondaryClass = secondary ? `mzp-t-secondary` : undefined;
+  $: darkClass = dark ? `mzp-t-dark` : undefined;
+  $: classSet = [
+    "mzp-c-button",
+    sizeClass,
+    productClass,
+    neutralClass,
+    darkClass,
+    secondaryClass,
+  ]
+    .filter((t) => t)
+    .join(" ");
+</script>
+
+<button use:forwardAll class={classSet} {disabled}><slot /></button>
--- a/src/components/forwardEvents.js
+++ b/src/components/forwardEvents.js
@@ -1,0 +1,70 @@
+// based on https://github.com/hperrin/svelte-material-ui/blob/master/packages/common/forwardEvents.js
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { bubble, listen } from "svelte/internal";
+
+export function forwardEvents(component, additionalEvents = []) {
+  const events = [
+    "focus",
+    "blur",
+    "fullscreenchange",
+    "fullscreenerror",
+    "scroll",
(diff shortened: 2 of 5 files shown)


src/components/Button.svelte
```diff
@@ -0,0 +1,31 @@
+<script>
+  import { get_current_component as getComponent } from "svelte/internal";
```
rhelmer, line 2: huh, odd that it's underscore_style not camelCaseStyle, looks like it's the only function like that in there... thanks for `as getComponent` :)

src/components/forwardEvents.js
```diff
@@ -0,0 +1,70 @@
+// based on https://github.com/hperrin/svelte-material-ui/blob/master/packages/common/forwardEvents.js
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { bubble, listen } from "svelte/internal";
+
+export function forwardEvents(component, additionalEvents = []) {
+  const events = [
```
rhelmer, line 6: Yikes :) Do you want a list of every event or just this particular list? There's a way to get the former but it's a little gnarly so this is fine.

rhelmer: Well I guess it's harder if you only want input-related events, I don't see a great alternative except https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/oninput but that's only for `input events on the <input>, <select>, and <textarea> elements. It also handles these events on elements where contenteditable or designMode are turned on.`

stories/protocol/ButtonView01.svelte
```diff
@@ -0,0 +1,42 @@
...
+      <Button {size}>standard</Button>
+      <Button {size} disabled>standard</Button>
+      <Button {size} secondary>secondary</Button>
+      <Button {size} secondary disabled>secondary</Button>
+      <Button {size} product>product</Button>
+      <Button {size} product disabled>product</Button>
+      <Button {size} product secondary>product / secondary</Button>
+      <Button {size} product secondary disabled>product / secondary</Button>
+      <Button {size} neutral>neutral</Button>
+      <Button {size} neutral disabled>neutral</Button>
```
rhelmer, line 39: Hm these all look the same size to me in Storybook is that expected? 
image

rhelmer (approved):
lgtm overall, qq about how this should look in storybook, the others are just commentary. Would you mind double-checking that the screenshot looks like what you are expecting before merging?

src/components/Button.svelte
hamilton: Yes, this is as svelte style difference. Lots of camel-case stuff for the deeper internals. Most other stuff is single-word though.

src/components/forwardEvents.js
hamilton: This is pretty standard and has not in practice caused any kind of bloat that I can tell. The Svelte Material UI library uses a similar thing. Open to thinking of another approach.

stories/protocol/ButtonView01.svelte
hamilton: hmm, that looks broken to me & isn't what I see. Will investigate.

hamilton: Fixed. my previous rebase was overzealous and this didn't break anything.

Issue #60, opened 2020-10-14
wait for study add-on to be installed before saying a study is "active" in the store


hamilton:
As of this filing, the UX around enrolling in a study is a little odd. The frontend should note that the study enrollment / add-on installation has finished only _after_ it has actually finished.

This will be much easier to fulfill after #46.

Dexterp37:
Note that this is related to #81 .

Dexterp37:
Closed by #97

PR #97, opened 2020-10-20
Send enrollment/deletion when addons lifetime changes

### LINKED ISSUE #81: cancelling study add-on install still shows study add-on as installed
I noticed that if I cancel extension install of a study add-on, then the core add-on `moz-extension://` control panel page shows the study as enrolled even though the study add-on is not actually installed. It still persists even after browser restart.

@hamilton @Dexterp37 Conceptually I think on startup the Ion Core should get a list of all add-ons on startup, and it should listen to add-on un/install and enable/disable events to update its internal state. It's ok to cache this state but we should make sure it does not get stale.

Dexterp37:
Fixes #81, #60.

This adds an `update-state` message sent by the background script to the UI. With this PR, the state is no longer duplicated between the UI storage and the background storage: the latter is the only source of truth.

This incorporate the work from @hamilton that updates the PR.

### DIFF
--- a/src/api/web-extension.js
+++ b/src/api/web-extension.js
@@ -76,9 +76,17 @@ export default {
   _connectionPort: null,
 
   // initialize the frontend's store from the add-on local storage.
-  async initialize(key) {
+  async initialize() {
+    // _stateChangeCallbacks holds all the callbacks we want to execute
+    // once the background sends a message with a new state.
+    this._stateChangeCallbacks = [];
+
+    // initialize the connection port.
     this._connectionPort =
       browser.runtime.connect({name: "ion-options-page"});
+
+    this._connectionPort.onMessage.addListener(
+      m => this._handleMessage(m));
 
     // The onDisconnect event is fired if there's no receiving
     // end or in case of any other error. Log an error and clear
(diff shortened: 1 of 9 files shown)


src/api/web-extension.js
```diff
@@ -174,4 +168,28 @@ export default {
...
+   *  type: "message-type",
+   *  data: { ... },
+   * }
+   * ```
+   */
+  async _handleMessage(message) {
+    switch (message.type) {
+      case "update-state": {
+        // TODO: Update the UI accordingly.
+        console.log('Received new studyes! ' + message.data);
```
Dexterp37, line 188: @hamilton this is your hook in point to update the UI!  See the docstrings for `_sendStateUpdateToUI` for the format of the `message.data` field for this message.

tests/core-addon/IonCore.test.js
```diff
@@ -45,6 +45,11 @@ describe('IonCore', function () {
     });
 
     this.ionCore = new IonCore();
+
```
hamilton, line 48: ah, this was what I was missing before – thanks!

hamilton (approved):
So thrilled this came together – thanks @Dexterp37! One last thing – we can remove `immer` as  dependency now that we've removed state updates in the frontend.

One thing to keep in mind as we go forward – we will have little bits of UX state we might want to keep in the store. Having a clear path to implementing & accounting for the schema as we go will be important. Either ajv or ts will help us a lot here. We've talked about it before, just want to note that this will save us energy in the future.

Feel free to merge at your discretion.

PR #104, opened 2020-10-22
Onboarding flow implemented in storybook

### LINKED ISSUE #84: Implement welcome page

### LINKED ISSUE #85: Implement terms page
- if the user clicks on "accept & participate", some event is fired from the terms page component.
- if the user clicks on "cancel", another event is fired from the terms page component.

### LINKED ISSUE #86: Implement demographics page

hamilton:
closes #84, closes #85, closes #86 

I found that a branch per page was getting a bit cumbersome, so this branch builds off of the commits for the welcome / TOS / demographics pages.


public/index.html
```diff
@@ -10,6 +10,7 @@
     <link rel="stylesheet" href="common.css" />
     <link rel="stylesheet" href="pioneer.css" />
     <link rel="stylesheet" href="global.css" />
+    <!-- <link rel="stylesheet" href="build/external.css" /> -->
```
rhelmer, line 13: Did you mean to leave this commented? Should it just be removed if so?

src/routes/terms-of-service/CallToAction.svelte
```diff
@@ -0,0 +1,65 @@
...
+  $: if (scrollY > 20) {
+    showArrow = false;
+  } else {
+    showArrow = true;
+  }
+
+  // let mounted = false;
+  // onMount(() => {
+  //   mounted = true;
+  // });
```
rhelmer, line 21: Same question re: commented code :) If it serves a purpose (for testing etc) then we should have a comment above it at least (or do it in different way preferably)

src/routes/terms-of-service/Content.svelte
```diff
@@ -0,0 +1,166 @@
...
+
+  <p>
+    Some sort of introduction lorem ipsum dolor sit amet, consectetur adipiscing
+    elit, sed do eiusmod tempor incididunt ut.
+  </p>
+
+  <hr />
+
+  <p>
+    Pioneer is an experimental initiative led by Mozilla to better understand
```
rhelmer, line 54: Should we just `s/Pioneer/Ion` on this page?

rhelmer (changes requested):
The onboarding storybook looks great!

One thing I noticed is that the study card looks a little off, missing CSS?

image

src/routes/terms-of-service/Content.svelte
hamilton: Yes, this copy will change in the very near future anyway.

hamilton:
@rhelmer thanks for the review! The issue with that study card story is I'm importing the _new_ css and not the old stuff. I've decided to just remove the study card story for now since it will be replaced before we even touch the current study card implementation.

hamilton:
@rhelmer Made the changes described above. I've done a pass on this with Ryan as well and have addressed his feedback. I think this is ready to merge.

PR #105, opened 2020-10-23
Enable the platform website to check for Core Addon installation

### LINKED ISSUE #103: Allow the Ion website to check for Ion Core addon presence
We exclusively want the website to know if the Core addon is installed. This should mitigate security implications for exposing any control to the web surface.

Given the minimal, read-only features we care about, can we do this without any privileged code?

Dexterp37:
Fixes #103 

This adds the required background and content script features in order to make the core addon communicate with the platform website.

### DIFF
--- a/core-addon/IonCore.js
+++ b/core-addon/IonCore.js
@@ -8,17 +8,19 @@ const DataCollection = require("./DataCollection.js");
 // The path of the embedded resource used to control Ion options.
 const ION_OPTIONS_PAGE_PATH = "public/index.html";
 
-// NOTE: if this URL ever changes, you will have to update the domain in
-// the permissions in manifest.json.
 const ION_DEFAULT_ARGS = {
-  availableStudiesURI: "https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/pioneer-study-addons-v1/records"
+  // NOTE: if this URL ever changes, you will have to update the domain in
+  // the permissions in manifest.json.
+  availableStudiesURI: "https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/pioneer-study-addons-v1/records",
+  website: "https://mozilla-ion.github.io",
 }
 
 module.exports = class IonCore {
    /**
   * @param {Object} args arguments passed in from the user.
-  * @param {String} args.availablStudiesURI the URI where the available Ion studies 
+  * @param {String} args.availableStudiesURI the URI where the available Ion studies
   *             information is listed.
+  * @param {String} args.website the URL of the platform website.
   */
   constructor(args = {}) {
     this._userArguments = {...ION_DEFAULT_ARGS, ...args};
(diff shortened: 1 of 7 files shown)


core-addon/IonCore.js
```diff
@@ -8,16 +8,17 @@ const DataCollection = require("./DataCollection.js");
...
+  // the permissions in manifest.json.
+  availableStudiesURI: "https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/pioneer-study-addons-v1/records",
+  website: "https://mozilla-ion.github.io",
 }
 
 module.exports = class IonCore {
    /**
   * @param {Object} args arguments passed in from the user.
-  * @param {String} args.availablStudiesURI the URI where the available Ion studies 
+  * @param {String} args.availablStudiesURI the URI where the available Ion studies
```
Dexterp37, line 21: TODO: missing docs

core-addon/content-script.js
```diff
@@ -0,0 +1,42 @@
...
+  let event = new CustomEvent(data.type, { detail: data });
+  window.dispatchEvent(data);
+}
+
+// Listen for messages from the page.
+// If the message was from the page script, show an alert.
+window.addEventListener("web-check", event => {
+  if (event.source === window) {
+    console.debug("Ion - 'web-check' message received from the page");
+    // TODO: use webextension-polyfill for chrome compatibility
```
Dexterp37, line 25: @hamilton is there any way for a single rollup configuration to generate two different artefacts?

We need to generate a `background.js` (which we do) and also to generate a `content-script.js`. We want to go through Rollup with this as we need to include the webextension-polyfill library.

hamilton: Rollup can create multiple bundles from one config. To achieve this, you [can export default an array of configs](https://rollupjs.org/guide/en/#configuration-files). Look in this section for a number of options.

Dexterp37: Thanks! That worked!

core-addon/IonCore.js
```diff
@@ -8,17 +8,19 @@ const DataCollection = require("./DataCollection.js");
...
```
rhelmer, line 21:
```suggestion
  * @param {String} args.availableStudiesURI the URI where the available Ion studies
```

core-addon/IonCore.js
```diff
@@ -232,6 +238,57 @@ module.exports = class IonCore {
...
+   *        the message.
+   * @returns {Promise} The response to the received message.
+   *          It can be resolved with a value that is sent to the
+   *          `sender` or rejected in case of errors.
+   */
+  _handleWebMessage(message, sender) {
+    console.log(`IonCore - received web message ${message} from ${sender}`);
+
+    // TODO: check for https, check for host, etc. Do something more
+    // appropriate by properly creating a `URL` object.
```
rhelmer, line 257: I think it's enough to create two URL objects and compare the `.origin`

rhelmer (approved):

core-addon/IonCore.js
Dexterp37: Oh, good suggestion. I updated this PR to cover it.

hamilton (approved):
lgtm. I think we'll probably want to possibly abstract out the website code as a js module so we can make sure future pages can be alerted to the add-on being installed.

Dexterp37:
> 
> 
> lgtm. I think we'll probably want to possibly abstract out the website code as a js module so we can make sure future pages can be alerted to the add-on being installed.

I created #113 to track this follow-up work.

PR #123, opened 2020-11-04
adds onboarding flow directly into the Core Add-On

### LINKED ISSUE #102: [meta] apply the onboarding UX into the Core Add-On
blocked by #84, #85, #86 

Once each of these pages are implemented, we're left with the task of integrating the page components into the Core Add-On. The work here looks like:

### **Welcome** 
-  the "Get Started" page moves the client forward. No state changes.
- the "How It Works" page should direct somewhere.

### **Terms**
- the "accept and participate" button should do a number of things:
  - the Core Add-On should create the needed user ID.
  - the Core Add-On should begin sending baseline telemetry.
  - the Core Add-On should move forward in the user flow to the Demographics page.
- the "cancel" button should direct back to Welcome page.
- closing the tab then reopening should direct user to the Welcome page.

### **Demographics**
- filling out the survey fully and submitting it should do the following:
  - the Core Add-On should save the survey results.
  - the page should direct to Current Studies.
   - the Core Add-On should note that the onboarding flow is now complete.
- **!!!** partially filling out the survey – _how should we handle this case?_
- clicking "skip" should direct to Current Studies & note that the user has gone through the onboarding flow in the storage.
- closing the tab then reopening should direct user to the Current Studies page.

hamilton:
closes #102 

This implements JUST the onboarding flow for the Core Add-On. Once we get past the onboarding step, it will do an ugly flicker while the stylesheets change back to the regular Ion-style card display.

This does _not_ submit the demographic survey nor does it save it in the add-on's storage. I have the basic storage schema directly in the [Demographics container component (take a look at this link)](https://github.com/mozilla-ion/ion-core-addon/blob/master/src/routes/demographics/Content.svelte#L57). This should be migrated outside of the component then imported into it in some way.

@Dexterp37 is this enough of a starting point to contribute to handling the demographic survey in storage / submitting it (#124)?

### DIFF
--- a/src/routes/Onboarding.svelte
+++ b/src/routes/Onboarding.svelte
@@ -1,4 +1,7 @@
 <script>
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
   import { onMount, getContext, createEventDispatcher } from "svelte";
   import Layout from "../components/Layout.svelte";
   import Main from "../components/Main.svelte";
--- a/src/routes/Onboarding.svelte
+++ b/src/routes/Onboarding.svelte
@@ -1,0 +1,81 @@
+<script>
+  import { onMount, getContext, createEventDispatcher } from "svelte";
+  import Layout from "../components/Layout.svelte";
+  import Main from "../components/Main.svelte";
+  import Welcome from "../routes/welcome/Content.svelte";
+  import Demographics from "./demographics/Content.svelte";
+  import TermsOfService from "./terms-of-service/Content.svelte";
+  import OnboardingCTAContainer from "../components/OnboardingCTAContainer.svelte";
+
+  import DemographicsCallToAction from "./demographics/CallToAction.svelte";
+  import TermsCallToAction from "./terms-of-service/CallToAction.svelte";
+
+  export let view = "welcome";
+  let mounted = false;
+
+  const store = getContext("rally:store");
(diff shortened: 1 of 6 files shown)


src/routes/Onboarding.svelte
```diff
@@ -0,0 +1,81 @@
...
+              send('demographics');
+            }}
+            on:cancel={() => {
+              send('welcome');
+            }} />
+        {:else if view === 'demographics'}
+          <DemographicsCallToAction
+            on:save={() => {
+              // Submit Demographics here.
+              // store.updateDemographicSurvey(results)
```
hamilton, line 72: @Dexterp37 Once we have a way to submit the demographic survey, this is where the store function would go.

Dexterp37: Ok, I can proceed as soon as I have `result` :)

Dexterp37:
> @Dexterp37 is this enough of a starting point to contribute to handling the demographic survey in storage / submitting it (#124)?

I don't think we need/should proceed in parallel on this. The UI work is pretty much unblocked: kindly make sure to provide some meaningful object containing the data that should be sent to the background script. Once this is done, I can pick it up.

Dexterp37 (commented):
@hamilton note that the "Age" question allows picking more than 1 answer. Is that intended?

hamilton:
@Dexterp37 that is an error – thanks for catching. It should be "single" instead of "multi". Pushed that change.

Dexterp37 (approved):

hamilton:
Thank you @Dexterp37 for the review!

PR #119, opened 2020-11-02
Add ESLint config

pdehaan:
Fixes #118

PR #125, opened 2020-11-04
Fix eslint problems

Dexterp37:
This follows up on #119 to fix the detected eslint problems (yay! it found a bug!).

This makes sure CI runs green, so next PRs can trust linting.


hamilton (approved):
This looks fine to me. I will need to go through our other eslint configs for svelte projects, since from experience I can say we're probably not doing enough there.

PR #134, opened 2020-11-04
Use assert module's strict mode

### LINKED ISSUE #121: assert.equal() is deprecated
Per, [Node API docs](https://nodejs.org/api/assert.html#assert_assert_equal_actual_expected_message) `assert.equal()` is deprecated since v9.9.0 &mdash; use `strictEqual()` instead.

```js
/** @deprecated since v9.9.0 - use strictEqual() instead. */
function equal(actual: any, expected: any, message?: string | Error): void;
```

Trying to find an ESLint config that can flag deprecated Node APIs for us. Early research shows https://github.com/mysticatea/eslint-plugin-node should do the trick, but I need to figure out how to configure it and silence some of the noisier recommended rules.

pdehaan:
Fixes #121

### DIFF
--- a/tests/core-addon/IonCore.test.js
+++ b/tests/core-addon/IonCore.test.js
@@ -426,8 +426,6 @@ describe('IonCore', function () {
 
   describe('_updateInstalledStudies()', function () {
     it('adds the ionInstalled property', async function () {
-      // We don't expect any update task to be running now.
-      assert.equal(this.ionCore._updateInstalledTask, undefined);
       // Kick off an update task.
       let studies =
         await this.ionCore._updateInstalledStudies(FAKE_STUDY_LIST);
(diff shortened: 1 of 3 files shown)


tests/core-addon/IonCore.test.js
```diff
@@ -427,7 +427,7 @@ describe('IonCore', function () {
   describe('_updateInstalledStudies()', function () {
     it('adds the ionInstalled property', async function () {
       // We don't expect any update task to be running now.
-      assert.equal(this.ionCore._updateInstalledTask, null);
+      assert.equal(this.ionCore._updateInstalledTask, undefined);
```
Dexterp37, line 430: AH! This is vestigial code, it's testing something that's no longer there :)
Would you kindly remove this line and the comment immediately above?

Dexterp37 (changes requested):
Thanks again for the nice PR!

tests/core-addon/IonCore.test.js
pdehaan: Removed.

Dexterp37 (approved):

tests/core-addon/IonCore.test.js
Dexterp37: Awesome, thank you so much!

PR #137, opened 2020-11-06
Add the MPL license header to all the relevant files in this repo

### LINKED ISSUE #136: Add the MPL license header to all the relevant files in this repo
For javascript files, this is the expected format:

```js
/* This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```

pdehaan:
Fixes #136 

Semi-confident that I got all the relevant .js and .svelte files.
I had to do some of the headers manually since it doesn't look like eslint-plugin-notice got all the .svelte files reliably, but I double checked with my <kbd>git grep</kbd> command below and verified that all the remaining results were either svg-in-Svelte files, or didn't have a `<script>` block (although we could add a `<script>` block with only a copyright header and hope it doesn't cause any downstream issues):

```sh
git grep -L "This Source Code Form is subject to the terms of the Mozilla Public" -- '*.js' '*.svelte'

src/components/MozillaRallyLogo.svelte
src/components/icons/Carat.svelte
src/components/icons/ExternalLink.svelte
src/copy/EnrollmentDisclaimer.svelte
src/copy/HowItWorks.svelte
src/copy/ValueProposition.svelte
src/copy/WhyItMatters.svelte
```


.eslintrc.js
```diff
@@ -57,8 +58,15 @@ module.exports = {
...
+    "notice/notice":[ "error", { mustMatch: "This Source Code Form is subject to the terms of the Mozilla Public", "templateFile": "copyright.txt" } ],
+
+    "eol-last": "warn",
+    "eqeqeq": "off", // TODO: "warn",
     "no-unused-vars": [ "error", { vars: "all", args: "none", ignoreRestSiblings: false } ],
     "no-var": "off", // TODO: "warn",
+    "no-warning-comments": "off", // TODO: "warn",
     "prefer-const": "off", // TODO: "warn",
+    "quotes": "off", // TODO: [ "error", "double" ],
+    "semi": [ "off", "always" ], // TODO: Set up prettier?
```
Dexterp37, line 70: Why do we need these?

.eslintrc.js
```diff
@@ -57,8 +58,15 @@ module.exports = {
     "node/no-unpublished-require": "off",
     "node/no-unsupported-features/es-syntax": "off",
 
+    "notice/notice":[ "error", { mustMatch: "This Source Code Form is subject to the terms of the Mozilla Public", "templateFile": "copyright.txt" } ],
+
+    "eol-last": "warn",
+    "eqeqeq": "off", // TODO: "warn",
     "no-unused-vars": [ "error", { vars: "all", args: "none", ignoreRestSiblings: false } ],
     "no-var": "off", // TODO: "warn",
+    "no-warning-comments": "off", // TODO: "warn",
```
Dexterp37, line 67: Why do we need this?

.eslintrc.js
```diff
@@ -57,8 +58,15 @@ module.exports = {
     "node/no-unpublished-require": "off",
     "node/no-unsupported-features/es-syntax": "off",
 
+    "notice/notice":[ "error", { mustMatch: "This Source Code Form is subject to the terms of the Mozilla Public", "templateFile": "copyright.txt" } ],
+
+    "eol-last": "warn",
+    "eqeqeq": "off", // TODO: "warn",
```
Dexterp37, line 64: Why do we need this?

.eslintrc.js
pdehaan: Not sure if we do, but i just noticed some inconsistency in the code.

If we do end up using Prettier, we can remove a bunch of the style-based ESLint rules in our config. But I don't have a lot of experience w/ Svelte to know how well it plays w/ Prettier so it might be more hassle than it's worth.

.eslintrc.js
pdehaan: Ah, I think this was more of a curiosity to see how many "TODO" comments we had throughout the code (and whether we should file issues for TODOs so we can track them better).
I figured ideally we'd eventually enable these rules, but they'd cause a semi-noisy diff currently so wasn't sure if we should bother or if we don't care [yet].

.eslintrc.js
Dexterp37: Got it! Thanks for explaining. Do you mind keeping this PR exclusively for the headers, leaving out all the other changes, since they might require discussion with the rest of the team? :)

.eslintrc.js
pdehaan: This is just a good rule in general since `===` is pretty rad.
But I don't have super strong feelings about many of these rules, so I can remove them for now or feel free to delete and merge if that's easier (since its midnight:48 locally and I probably won't be online much longer).

Dexterp37 (changes requested):
Let's keep the changes in this PR exclusively for the headers (and, at most the eol-last change!), leaving out all the other changes, since they might require discussion with the rest of the team. :-)

Dexterp37 (approved):
That was quick! Thank you, once again!

PR #190, opened 2020-11-20
Remove the "management" permission from the study template

Dexterp37:
This removes the need of the "management" permission to check if the Core Add-on is installed from the studies.

Fixes mozilla-ion/ion-basic-study#14

TODO:

- [x] Add test coverage.
- [x] Perform manual QA.
- [ ] Release a new package version.
- [ ] Bump the version in the study template.


support/rally.js
```diff
@@ -55,10 +55,22 @@ module.exports = class Rally {
...
+        type: "core-check",
+        data: {}
+      }
+      let response =
+        await browser.runtime.sendMessage(CORE_ADDON_ID, msg, {});
+
+      if (!response
+          || response.type !== "core-check-response"
+          || response.data.enrolled !== true) {
+        return Promise.reject(
```
rhelmer, line 68: Nit, but is there a difference between `return Promise.reject()` and `throw` since this is an `async` function?

support/rally.js
```diff
@@ -55,10 +55,22 @@ module.exports = class Rally {
...
+          new Error(`Rally._checkRallyCore - unexpected response returned ${response}`));
+      }
     } catch (ex) {
-      console.error("Rally._checkRallyCore - core addon not found", ex);
-      return Promise.reject(ex);
+      return Promise.reject(
```
rhelmer, line 72: Same q as above, wondering if `async` sugars this away.

core-addon/IonCore.js
```diff
@@ -222,6 +222,15 @@ module.exports = class IonCore {
     }
 
     switch (message.type) {
+      case "core-check": {
+        let enrolled = !!(await this._storage.getIonID());
+        return Promise.resolve({
```
rhelmer, line 227: nit, does returning `Promise.resolve()` from an `async` function do anything different than just `return`? I think async/await just does syntactic sugar over this and auto-wrap/unwrap of promises so I don't see the difference testing this out locally.

rhelmer (approved):

core-addon/IonCore.js
Dexterp37: You're right! Fixing this.

Issue #66, opened 2020-10-15
Merge back build system changes from mozilla-extensions clone


rhelmer:
I've made a mess of the build system in the mozilla-extensions clone (required for privileged signing), I need to clean that up and merge it back https://github.com/mozilla-extensions/ion-core-addon

Let's file bugs upstream also on the following issues:

- [ ] dep-signing task does not work for forks
- [ ] build target should be configurable
- [ ] determine the problem with the dev dependencies in our `package.json`

I believe that we need to merge https://github.com/mozilla-ion/ion-core-addon/pull/199 in order to get TC to start paying attention to this repo. We can further debug the latter two issues after that's merged.

rhelmer:
Testing this in https://github.com/mozilla-extensions/ion-core-addon/pull/1 - since these are separate repos and not forks, one must fork the mozilla-extensions/ion-core-addon separately which is annoying but OK for now (we can't use a fork because there's a bug in addons-pipeline where dep-signed builds will not be produced)

Dexterp37:
> 
> 
> Testing this in [mozilla-extensions/ion-core-addon#1](https://github.com/mozilla-extensions/ion-core-addon/pull/1) - since these are separate repos and not forks, one must fork the mozilla-extensions/ion-core-addon separately which is annoying but OK for now (we can't use a fork because there's a bug in addons-pipeline where dep-signed builds will not be produced)

Thanks for doing this! I have a few thoughts, but I think I'm missing something, so bear with me as I might be saying many wrong things in this comment :-)

mozilla-extensions/ion-core-addon#1 seems to be trying to merge from [your fork](https://github.com/rhelmer/ion-core-addon-1) to the fork in `mozilla-extensions`, which we require for signing. A few questions below:

* Can we treat the fork in `mozilla-extensions` as "read-only"? This means any chance that needs to happen will be required to happen in `mozilla-ion/ion-core-addon`. Once merged, we can open a PR directly in `mozilla-extensions/ion-core-addon` updating it to the latest `master` of our main repo. This should always be a fast-forward merge, with no conflicts. Ideally, this could be automated: if we cut a new release in this repo (through GitHub release tags), we could open that PR automatically and trigger any dependant signing job. How does that sound?
* Are we lacking any tweak that was made to `mozilla-extensions/ion-core-addon` and is not in `mozilla-ion/ion-core-addon`?

@rhelmer what do you think of the above?

rhelmer:
> 
> 
[...]

PR #197, opened 2020-11-20
Build system changes for mozext repo, fixes issue #66

rhelmer:
Sorry for the mess :) There are a few problems here, and they are all driven by complications on the taskcluster side:

- I could use help cleaning up the build target. I'm not sure if we can have the taskcluster CI run anything other than `build` so we may have to have this target generate the add-on. I will follow up on this.
- taskcluster CI seems to have a problem with our dev dependencies; it actually calls `yarn build` so I've added a `yarn.lock`, this is a pain to maintain and I don't think we should switch to `yarn` generally. Following up on this per above.


Dexterp37:
@rhelmer how does this relate to #199 ?

rhelmer:
Sorry this is a dupe of PR #199

PR #199, opened 2020-11-21
Build system changes for mozext repo, for issue #66

rhelmer:


package.json
```diff
@@ -54,7 +54,8 @@
   },
   "dependencies": {
     "sirv-cli": "^1.0.0",
-    "web-ext": "^5.1.0"
+    "web-ext": "^5.1.0",
+    "yarn": "^1.22.10"
```
pdehaan, line 58: Did you mean to include yarn as a dep here?

rhelmer: Thanks, yeah unfortunately the Taskcluster CI on mozilla-extensions/ org seems to prefer `yarn` and it's have some problem with the dev dependencies in our `package.json`. I suspect this is just a bug, we need to figure where to file this (and also the bug about the dep-signing job not working for forks, but that might be harder because of the security assumptions TC seems to be set up with).

rhelmer: Using `yarn.lock` was suggested as a workaround, but I'd be really happy if we can just get rid of it.

rhelmer:
OK here's what I am going to suggest. Pre https://github.com/mozilla-ion/ion-core-addon/issues/66#issuecomment-731507085 what I think is going on here is that taskcluster requires the `.taskcluster.yml` to be on the master branch or it won't run at all, so we can't test further changes in PRs.

I suggest we merge this patch (maybe clean up the build targets first, but the TC automation seems to be hardcoded to `yarn build` which isn't great) so we can push it over to the mozilla-extensions/ion-core-addon repo and then be able to use PRs.

@Dexterp37 wdyt?

package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "svelte-app",
-  "version": "1.0.0",
+  "name": "ion_core",
+  "version": "2.2.0",
   "scripts": {
-    "build": "rollup -c",
+    "build": "rollup -c && rollup -c rollup.config.addon.js --config-study-list-url=/public/locally-available-studies.json && web-ext build --overwrite-dest && mv web-ext-artifacts/*.zip web-ext-artifacts/ion_core.xpi",
```
Dexterp37, line 5: @rhelmer why are you changing this command?

Do we need to add another command to package a release? In the study template, for example, you recommended adding the [`package` command](https://github.com/mozilla-ion/ion-basic-study/blob/75249cd338da6b10cc7e38e7ff72d290323b18da/package.json#L7).

package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "svelte-app",
-  "version": "1.0.0",
+  "name": "ion_core",
+  "version": "2.2.0",
```
Dexterp37, line 3: Since we now have two places in which to change the version, we should probably:

- add a doc in the docs/ directory to document the release process;
- create a bash script to prepare the repo to cut a new release (i.e. bump the version in all the places we need to bump it in).

We did something similar in Glean: [script](https://github.com/mozilla/glean/blob/main/bin/prepare-release.sh), [docs](https://mozilla.github.io/glean/book/dev/cut-a-new-release.html).

Note that this is fine to do as a separate issue.

Dexterp37 (changes requested):
I'm fine with landing the YML part here. I have smaller concerns with the other parts, see my comments below.

Dexterp37:
> I suggest we merge this patch (maybe clean up the build targets first, but the TC automation seems to be hardcoded to `yarn build` which isn't great) so we can push it over to the mozilla-extensions/ion-core-addon repo and then be able to use PRs.
> 
> @Dexterp37 wdyt?

Yes, I think that's a good idea! Can we just clear up the lockfile before we do that?

Dexterp37 (approved):
Thanks Rob! Let's revert the package-lock.json file (get rid of all the dependecy version changes) and merge this!

package.json
rhelmer: There's a slack thread I started in #addons-pipeline about this, basically the `build` target is hardcoded in the taskcluster CI AFAIK :( Trying to figure this out though I am seeing some additional runs for some of our other targets so this needs investigation.

I *think* what the CI does is run `yarn build`, so we might just be seeing the effect of whatever that does. I don't think we should switch to `yarn` but we need to support it somewhat for now I guess.

package.json
rhelmer: We do have the start of docs for the first point https://github.com/mozilla-ion/ion-core-addon/blob/master/docs/ION_CORE_PUBLISHING.md

The second is a good idea, since we want to unblock this issue ASAP I'm going to file a separate issue for this as you suggest.

rhelmer:
> 
> 
> Thanks Rob! Let's revert the package-lock.json file (get rid of all the dependecy version changes) and merge this!

Thanks! I am going to run `npm ci` as you suggest and will push the result of that.

rhelmer:
> 
> 
> > Thanks Rob! Let's revert the package-lock.json file (get rid of all the dependecy version changes) and merge this!
> 
> Thanks! I am going to run `npm ci` as you suggest and will push the result of that.

Hm, this produces no changes... does our `package-lock.json` just need to be updated? I'm going to merge this so we can move forward but happy to help untangle if this is not right!

rhelmer:
OK I pushed `master` over to https://github.com/mozilla-extensions/ion-core-addon (no force-push needed) and CI is running there now, so we can make future changes via PRs.

package.json
rhelmer: > 
> 
> Since we now have two places in which to change the version, we should probably:
> 
>     * add a doc in the docs/ directory to document the release process;
> 
>     * create a bash script to prepare the repo to cut a new release (i.e. bump the version in all the places we need to bump it in).
> 
> 
> We did something similar in Glean: [script](https://github.com/mozilla/glean/blob/main/bin/prepare-release.sh), [docs](https://mozilla.github.io/glean/book/dev/cut-a-new-release.html).
> 
> Note that this is fine to do as a separate issue.

Filed issue #210

Issue #195, opened 2020-11-20
release version 2.3 of the core add-on


rhelmer:
Let's cut a new release on github, and set it up on https://mozilla-ion.github.io/ion-core-addon/

We can put the dev-signed build up for now, and also use this issue to track getting it prod-signed.

rhelmer:
Release is done https://github.com/mozilla-ion/ion-core-addon/releases/tag/v2.2 - I am leaving this issue open to track the following:

- [ ] post dev-signed XPI to the above release
- [ ] point https://mozilla-ion.github.io/ion-core-addon/ at dev-signed add-on
- [ ] get a production-signed XPI from AMO admins and update release + https://mozilla-ion.github.io/ion-core-addon/

rhelmer:
Actually I think we should do a 2.2.1 to include `mozilla-recommendation.json` per https://github.com/mozilla-ion/ion-core-addon/issues/159#issuecomment-731624422 so we get "badging" and more importantly suppresses the warning prompt when installing from a non-AMO origin.

Dexterp37:
@rhelmer can we release the core addon with a pre-1.0 version? It seems a bit weird to be at version 2 when 1 was never out :E

rhelmer:
This is taking long enough that we're on 2.3 https://github.com/mozilla-ion/ion-core-addon/releases/tag/v2.3

> @rhelmer can we release the core addon with a pre-1.0 version? It seems a bit weird to be at version 2 when 1 was never out :E

I'm not sure why we started with 2.0 :) @Dexterp37 would you mind filing an issue? Since it's pre-release I suggest we clean up the existing tags and reset to 0.23

hamilton:
Closing this one for now, and @rhelmer is going to open an issue for 0.6

PR #202, opened 2020-11-21
Add recommendation file, suppress install warning

### LINKED ISSUE #159: Provide a signed version of the Core Addon; potentially enable “Badging" (CA-P0, CA-P1)
Provide a signed version of the Core Addon
- And potentially enable “Badging"

rhelmer:
See issue #195, closes #159. This implements "add-on badges",
which also removes the warning prompt when starting an install
from an untrusted origin. See https://blog.mozilla.org/addons/2020/10/05/new-add-on-badges/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,9 +60,6 @@ jobs:
           name: Install the dependencies
           command: npm install && sudo add-apt-repository ppa:ubuntu-mozilla-daily/ppa && sudo apt update && sudo apt install firefox-trunk
       - run:
-          name: Download and unpack GeckoDriver for Selenium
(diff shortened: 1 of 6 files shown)


rhelmer:
Note that I haven't actually tested this yet, I could use help doing so :) From reading the addons manager code I don't think temporarily-installed is good enough, disabling signature checks might be. Using dev-root should be (but this requires us to have our tasckluster CI wired up which we're working on elsewhere)

Dexterp37:
@rhelmer I believe some Add-on peer should review this code. Also note that web-ext linting is calling this out as an error, too :)

Dexterp37:
@rhelmer can this be closed?

rhelmer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1681331 just landed on Nightly which suppresses this prompt for privileged extensions, I don't think we need to figure out recommendation badges for the privileged signing infra right now.

Issue #55, opened 2020-10-14
Run integration tests on CI


Dexterp37:
In #16 we added running unit test coverage and linting for the addon on CircleCI.

We now need a way to run integration tests using real browsers (headless?).

rhelmer:
I've been looking at what the major extensions like uBlock and AdBlockPlus and others do, they seem to have automation for building the extension and running unit tests but I'm not finding a ton on integration tests.

The Lighthouse extension from Google does this, and uses Puppeteer with headless Chrome: https://github.com/GoogleChrome/lighthouse/pull/4640

There are various examples of using Mocha with Puppeteer and Selenium (I'm still trying to understand the differences, but it looks like Puppeteer might be easier to set up and provide nicer integration into various browser features so it's probably worth seriously considering since it supports Firefox as well (which was one of the big sticking points for using Selenium for a while)

rhelmer:
OK here's an example of using Puppeteer with `web-ext`:

```
const path = require("path");
const puppeteer = require("puppeteer");
const webExt = require("web-ext");

async function run() {
  webExt.util.logger.consoleStream.makeVerbose();
  const PORT = 1234;
  const extensionPath = "./";
  const runner = await webExt.cmd.run(
    {
      sourceDir: path.join(__dirname, extensionPath),
      firefox: {
        darwin: "/Applications/Firefox Nightly.app/Contents/MacOS/firefox-bin",
        win32: `${process.env.PROGRAMFILES}\\Firefox Nightly\\firefox.exe`,
      }[process.platform],
      args: [`--remote-debugging-port=${PORT}`],
    },
    {
      shouldExitProgram: false,
    }
  );
  const { debuggerPort } = runner.extensionRunners[0].runningInfo;
  console.log(debuggerPort);
  return await puppeteer.connect({
    browserWSEndpoint: `ws://localhost:${PORT}`,
  });
}

run()
  .then((result) => console.debug("running"))
  .catch((err) => console.error(err));
```

The `puppeteer.connect()` bit allows us to then use the Puppeteer API which is now hooked into the Firefox devtools:
https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#puppeteerconnectoptions

[...]

PR #212, opened 2020-12-02
Firefox integration testing

rhelmer:
Fixes issue #55


rhelmer:
OK this is rough :) But it demonstrates that we can run mochitest headless in circle CI, you can see a mochitest running (with one failure) here: https://app.circleci.com/pipelines/github/mozilla-ion/ion-core-addon/239/workflows/bc311e8a-f4b0-4c50-87b9-6670afd15967/jobs/707/parallel-runs/0/steps/0-106

This fully clones the mozilla-central repo on every run, and it occasionally runs out of memory since it's a medium-sized executor :). This means that when it does work the clone part is very slow, and the `bootstrap` takes too much time as well: https://app.circleci.com/pipelines/github/mozilla-ion/ion-core-addon/239/workflows/bc311e8a-f4b0-4c50-87b9-6670afd15967/jobs/707

CircleCI supports caching assets though, so this might be a way to make the performance acceptable.

@Dexterp37 wdyt? I think we can get something going for Firefox this way, I don't think Chrome provides a great way for us to easily use the browser UI though. So for Chrome I think we will need to use something like puppeteer and accept that everything is loaded as temporary extensions etc. We should be able to get excellent test coverage for Firefox though, and from my investigations into puppeteer and selenium I think that we won't get perfectly cross-browser tests in any case.

Dexterp37:
> This fully clones the mozilla-central repo on every run, and it occasionally runs out of memory since it's a medium-sized executor :). This means that when it does work the clone part is very slow, and the `bootstrap` takes too much time as well: https://app.circleci.com/pipelines/github/mozilla-ion/ion-core-addon/239/workflows/bc311e8a-f4b0-4c50-87b9-6670afd15967/jobs/707

Why do we need to clone m-c in the first place? Just to run mochitests? Skimming through MDN there's maybe a way that does not require cloning m-c. [See `Running tests without mach`](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Mochitest).

> @Dexterp37 wdyt? 

This is a good approach! Let's leave other browsers aside for now.

rhelmer:
> 
> 
> > This fully clones the mozilla-central repo on every run, and it occasionally runs out of memory since it's a medium-sized executor :). This means that when it does work the clone part is very slow, and the `bootstrap` takes too much time as well: https://app.circleci.com/pipelines/github/mozilla-ion/ion-core-addon/239/workflows/bc311e8a-f4b0-4c50-87b9-6670afd15967/jobs/707
> 
> Why do we need to clone m-c in the first place? Just to run mochitests? Skimming through MDN there's maybe a way that does not require cloning m-c. [See `Running tests without mach`](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Mochitest).

Hm, given the warnings in that doc about using marionette and your conversation with test automation folks, I'm going to see if I can switch this over to Selenium and use (hopefully) also use a nicer test runner too.

rhelmer:
@Dexterp37 OK this is ready for review! It's the most minimal example I can think of, essentially it boils down to:

1. download geckodriver from https://github.com/mozilla/geckodriver/releases and put on PATH
2.
```bash
npm run integration-test
```

This will run a headless browser instance of Firefox Nightly, which installs the local build of the core add-on and installs it. It does this by disabling signature checks and then clicking on the add-on install prompts.

We obviously need more tests than this :) But this alone would catch bugs such as https://github.com/mozilla-ion/ion-core-addon/pull/211 and we can start building on it.

Thanks for following up with whimboo about this, I think with the Firefox chrome context we will be able to do the sort of tests we want without Marionette and Mochitest etc.

package.json
```diff
@@ -15,7 +15,8 @@
     "lint-css": "stylelint 'public/*.css' '.storybook/*.css' 'src/**/*.svelte' 'stories/**/*.svelte'",
     "storybook": "start-storybook -p 6006 -s ./",
     "build-storybook": "build-storybook -s ./",
-    "test": "npm run test-addon && npm run test-support-library"
+    "test": "npm run test-addon && npm run test-support-library",
+    "integration-test": "npm build && node tests/core-addon/selenium/core_addon.js"
```
Dexterp37, line 19: Can we retain the same style of the other test commands? e.g. `test-integration`?

Please also note that this command is failing on CI:

> npm WARN build `npm build` called with no arguments. Did you mean to `npm run-script build`?

It should be `npm run build && ...`

tests/core-addon/selenium/core_addon.js
```diff
@@ -0,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
Dexterp37, line 1: Instead of calling the directory `selenium`, can we rather call it `integration`? It might be beneficial to also move the unit tests to a `unit` directory.

package.json
Dexterp37, line 19: Can we also use the `mocha` test runner as we do for the other tests?

tests/core-addon/selenium/core_addon.js
```diff
@@ -0,0 +1,37 @@
...
+// This is the path to Firefox Nightly on Ubuntu with the Mozilla PPA
+if (process.platform === "linux") {
+  firefoxOptions.setBinary("/usr/bin/firefox-trunk");
+}
+const driver = new Builder()
+  .forBrowser("firefox")
+  .setFirefoxOptions(firefoxOptions)
+  .build();
+
+(async function install() {
```
Dexterp37, line 22: We want to use the Mocha runner for tests. Can you reshape this to to work with Mocha? Here's an example: https://github.com/mochajs/mocha-examples/blob/master/packages/selenium/test/selenium.spec.js

Dexterp37 (changes requested):

package.json
rhelmer: oops I messed that latter part up moving from circle config to our `package.json`, sorry about that :/ corrected the name too.

tests/core-addon/integration/core_addon.js
```diff
@@ -0,0 +1,44 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const firefox = require("selenium-webdriver/firefox");
+const { Builder, By, until } = require("selenium-webdriver");
+const assert = require("assert");
+
+let driver = null;
```
Dexterp37, line 9: I think this can be safely removed...
```suggestion
```

tests/core-addon/integration/core_addon.js
```diff
@@ -0,0 +1,44 @@
...
+
+// This is the path to Firefox Nightly on Ubuntu with the Mozilla PPA
+if (process.platform === "linux") {
+  firefoxOptions.setBinary("/usr/bin/firefox-trunk");
+}
+
+describe("Selenium", function () {
+  // eslint-disable-next-line mocha/no-hooks-for-single-case
+  beforeEach(async function () {
+    driver = await new Builder()
```
Dexterp37, line 23:
```suggestion
    this.driver = await new Builder()
```
... and we could do this

tests/core-addon/integration/core_addon.js
```diff
@@ -0,0 +1,44 @@
...
+      .forBrowser("firefox")
+      .setFirefoxOptions(firefoxOptions)
+      .build();
+  });
+
+  // eslint-disable-next-line mocha/no-hooks-for-single-case
+  afterEach(async function () {
+    await driver.quit();
```
Dexterp37, line 31:
```suggestion
    await this.driver.quit();
```

tests/core-addon/integration/core_addon.js
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  it("should install the core add-on", async function () {
+    await driver.get(`file:///${__dirname}/index.html`);
+    await driver.wait(until.titleIs("Installation Test"), 1000);
+    await driver.findElement(By.id("install")).click();
+    // switch to browser UI context, to interact with Firefox add-on install prompts.
+    await driver.setContext(firefox.Context.CHROME);
+    await driver.findElement(By.css(`[label="Add"]`)).click();
+    const result = await driver.findElement(By.css(`[label="Okay, Got It"]`));
```
Dexterp37, line 41:
```suggestion
    await this.driver.get(`file:///${__dirname}/index.html`);
    await this.driver.wait(until.titleIs("Installation Test"), 1000);
    await this.driver.findElement(By.id("install")).click();
    // switch to browser UI context, to interact with Firefox add-on install prompts.
    await this.driver.setContext(firefox.Context.CHROME);
    await this.driver.findElement(By.css(`[label="Add"]`)).click();
    const result = await this.driver.findElement(By.css(`[label="Okay, Got It"]`));

```

Dexterp37 (approved):
r+ with some optional nits

tests/core-addon/integration/core_addon.js
rhelmer: Hm, this is there because `describe` sets up and tears it down so it's created in a different scope (`beforeEach` function). I can just make it `let driver;` and move into the `describe` function I guess?

Issue #214, opened 2020-12-02
Allow non-addon `package.json` in addons repo


Dexterp37:
Our signing PR on the mozilla-extensions forked repo [here](https://github.com/mozilla-extensions/ion-core-addon/pull/5) fails to sign our addon. That's because the [TaskCluster configuration](https://github.com/mozilla-extensions/ion-core-addon/blob/0036c7888648633499951ac45991e866b6ec92ec/.taskcluster.yml#L20) uses a template which looks for `package.json` in [all the directories](https://github.com/mozilla-extensions/xpi-template/blob/bb609c7929e65c31803439a5872aef05260d0b96/taskcluster/xpi_taskgraph/xpi_manifest.py#L60-L87).

It assumes they are addons (in our case they are just NPM libraries we publish) and looks for artefacts to sign.

We talked to aki from the Addons team, we agreed that we could provide a PR for the template that allows looking for a `dontsign` file in the directory where the `package.json` file is. If found, the template should not attempt to sign the artefact.

Dexterp37:
This should probably be fixed by mozilla-extensions/xpi-template#19

Dexterp37:
It was fixed. This is done!

PR #215, opened 2020-12-02
Fix taskcluster ci failures

rhelmer:
This is equivalent to the PR on https://github.com/mozilla-extensions/ion-core-addon/pull/5 - note that the support library is causing us some trouble over there because the taskcluster signing automation stuff searches recursively for `package.json` files and assumes anything it finds is an extension that produces a `.xpi` :/

@hamilton I think @Dexterp37 is done with his day, would you mind taking a look so we can merge this and move signing forward for now? Thanks!


rhelmer:
Oh, and issue #214 is where we are tracking actually fixing the support library issue.

Dexterp37 (approved):

Issue #217, opened 2020-12-03
Use pre-1.0 release versions


rhelmer:
At some point we bumped our `manifest.json` up to `2.0`, @Dexterp37 points out that this doesn't make very much sense though.

Should we rename the releases and tags on this repo? Since we haven't done a real production release yet I feel like we have more leeway to make weird changes like this.

We've done 5 releases on https://github.com/mozilla-ion/ion-core-addon/releases - wdyt about renaming the tags to be `0.1.0`...`0.5.0` and then calling what we have on `main` now `0.6.0`?

rhelmer:
@hamilton might be interested too.

Dexterp37:
Fixed by #218

Dexterp37:
Reopening: we still need to either [delete the old tags](https://docs.github.com/en/enterprise/2.16/user/github/administering-a-repository/editing-and-deleting-releases#deleting-a-release) or rename them. I'd go for deleting them, which seems to be easier.

Dexterp37:
> 
> 
> Reopening: we still need to either [delete the old tags](https://docs.github.com/en/enterprise/2.16/user/github/administering-a-repository/editing-and-deleting-releases#deleting-a-release) or rename them. I'd go for deleting them, which seems to be easier.

I cleaned up our releases [following this process](https://github.community/t/how-to-delete-release-completely/1466/2) and removed all the tags.

PR #218, opened 2020-12-03
use consistent version numbers and reset to 0.6.0, addresses part of …

rhelmer:
…issue #217


Dexterp37 (approved):
Thanks for doing this!

PR #225, opened 2020-12-04
Uninstall the core addon when leaving Rally

### LINKED ISSUE #207: update _unenroll in IonCore.js to also uninstall the Core Add-On once all the cleanup is done
This is related to #205 – when a user has decided to leave Rally, the `_unenroll` function in `IonCore.js` should also (and aside from the deletion pings, sending uninstall messages to studies, etc.):

1. close all active rally extension URLs 
2. uninstall the Core Add-On
3. (optional) If we can get to a minimal "sorry to see you go!" website page together before GTM, we can open it for the user and make it the active tab before uninstalling the Core Add-On

Dexterp37:
This fixes #207 .

### DIFF
--- a/core-addon/IonCore.js
+++ b/core-addon/IonCore.js
@@ -188,8 +188,7 @@ module.exports = class IonCore {
         // from the control panel.
         return this._unenrollStudy(message.data.studyID);
       case "unenrollment":
-        return this._unenroll()
-                   .then(r => this._sendStateUpdateToUI());
+        return this._unenroll();
       case "update-demographics":
         return this._updateDemographics(message.data);
       default:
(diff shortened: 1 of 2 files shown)


rhelmer (approved):
lgtm! What do you think about adding an integration test too?

rhelmer:
> 
> 
> lgtm! What do you think about adding an integration test too?

On further thought, I suggest that we instead add an un/enrollment test that goes over the whole install flow, which can include a check that the extension is uninstalled. Then in the future we can either modify the un/enrollment test or add new tests as necessary, but we should get that basic groundwork in place first.

I filed issue #227 and I have a patch in progress which will cover the current PR.

PR #247, opened 2020-12-09
point to latest 0.6.0 version on taskcluster

rhelmer:
@Dexterp37 cleaned up our existing releases (thanks!) but unfortunately this breaks mozilla-ion.github.io/ion-core-addon/ as the 0.6.0 release is no longer present.

Hamilton, would you mind merging this temporarily? It points to the 0.6.0 release that's pending sign-off, I'd like to be able to show the currently enrollment flow to add-ons folks today so they know what I'm referring to.


rhelmer:
Actually let's scrub this for now, and do a proper release tomorrow. AMO folks understood what I meant we don't need to demo it to them today.

PR #248, opened 2020-12-09
Link GitHub Pages site to 0.6.1 pre-release

rhelmer:
OK I re-did this more properly - @Dexterp37 since we don't have the signed add-on yet, I took the latest 0.6.1 release from your PR on mozilla-extensions/ion-core-addon and added it as an artifact to our corresponding GH release.

In the future we should not need to do this, we will go with the signed add-on instead. But for today I want to unbust our GH pages site in a reasonable way.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -200,7 +200,7 @@
     <a
       id="installButton"
       tabindex="0"
-      href="https://github.com/mozilla-ion/ion-core-addon/releases/download/v2.3/ion_core.xpi"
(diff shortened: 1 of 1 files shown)


rhelmer:
I feel like it's important to keep our GH pages site as functional as we can, which is why I am being a pest about this :)

hamilton (approved):
lgtm

Dexterp37:
> 
> 
> I feel like it's important to keep our GH pages site as functional as we can, which is why I am being a pest about this :)

It is, good call.

PR #269, opened 2020-12-16
Allow using the Rally.js support library without a bundler.

Dexterp37:
This is preparatory work to use rally.js within integration tests.

TODO:

- [x] Perform a manual QA round.

Checklist for reviewer:

### DIFF
--- a/support/rally.js
+++ b/support/rally.js
@@ -5,7 +5,7 @@
 const CORE_ADDON_ID = "ion-core-addon@mozilla.org";
 const SIGNUP_URL = "https://mozilla-rally.github.io/core-addon/";
 
-module.exports = class Rally {
+class Rally {
   /**
    * Initialize the Rally library.
    *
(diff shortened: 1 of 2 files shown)


Dexterp37:
I tested this manually, everything works just fine. This is good to review @rhelmer !

support/rally.js
```diff
@@ -179,3 +179,8 @@ module.exports = class Rally {
     }
   }
 }
+
+// Make this library Require-able.
```
brizental, line 183: I believe you can change your output format from `iife` to `umd` and then you don't need to do this.

brizental (approved):
r+wc

support/rally.js
Dexterp37: Thanks, as we talked on Matrix we won't be always using a bundler (e.g. tests) so that won't help :(

PR #271, opened 2020-12-17
Rename `ion` to `rally` in the codebase

### LINKED ISSUE #114: Rename "ion" to "rally" in codebase
This is to rename things within the core addon to "ion". This should be a simple text search + rename.

Dexterp37:
Fixes #114

Checklist for reviewer:


rhelmer:
I can still find some mentions of Ion in the codebase:
```zsh
$ grep --exclude-dir=.git --exclude-dir=node_modules -r Ion .
./src/App.svelte:          store.updateIonEnrollment(false); 
./src/create-store.js:        `Ion - changing study ${studyID} enrollment to ${coercedEnroll}`);
[...]
```

Also there is the issue of the add-on ID itself which is `ion-core-addon@mozilla.org`

I don't think will be harmful to merge @Dexterp37 what do you think about the above? Changing the add-on ID I don't think will be super disruptive, so if we're going to it'd be good to do it now before 1.0 :)

rhelmer (changes requested):
Sorry I left this as a comment instead of a review:

1. I see some instances of Ion still do we care about these? Both in function names and comments.
2. Do we want to change the add-on ID ("ion-core-addon@mozilla.org")? If we do we should probably drop the `addon` bit since it's redundant, given that this is an add-on ID :) Maybe "rally-core@mozilla.org"?

rhelmer (approved):
Actually I am going to approve since nothing is wrong, there are just some other parts you might want to change too - however at minimum you're going to want the `package.json` build target and name etc. all consistent, in order to fix https://github.com/mozilla-extensions/xpi-manifest/pull/92

Dexterp37:
> 
> 
> Sorry I left this as a comment instead of a review:
> 
>     1. I see some instances of Ion still do we care about these? Both in function names and comments.
> 
>     2. Do we want to change the add-on ID ("[ion-core-addon@mozilla.org](mailto:ion-core-addon@mozilla.org)")? If we do we should probably drop the `addon` bit since it's redundant, given that this is an add-on ID :) Maybe "[rally-core@mozilla.org](mailto:rally-core@mozilla.org)"?

Whoops, sorry. I ended up doing the full rename here to avoid any confusion. Thank you for spotting these.

Dexterp37:
@rhelmer unfortunately integration test will remain broken until we cut a new release of the Rally.js and change the study template to use it, then cut a new release :)

Dexterp37:
Note: this will require mozilla-extensions/xpi-manifest#95 for us to be able to cut a new release.

rhelmer (approved):

rhelmer:
Well changes lgtm but the integration test disagrees :)

rhelmer:
> Well changes lgtm but the integration test disagrees :)

I suspect it's https://github.com/mozilla-rally/core-addon/blob/master/tests/core-addon/integration/core_addon.js#L82 at minimum

Dexterp37:
> 
> 
> > Well changes lgtm but the integration test disagrees :)
> 
> I suspect it's https://github.com/mozilla-rally/core-addon/blob/master/tests/core-addon/integration/core_addon.js#L82 at minimum

Yeah, that's changed in this PR. The reason why it's breaking is because the `rally.js` version the current "ion-basic-study" is using talks to the wrong add-on id (because we changed the id in this PR).

PR #280, opened 2020-12-19
Do not hide study addons closes issue 235

### LINKED ISSUE #235: study add-ons are hidden from UI on options page reload
If you run the core add-on locally (`web-ext` is fine) and enroll then navigate to the available studies, they are displayed and can be joined/unjoined.

However if you reload the options page (close tab and hit icon, or just press reload in the browser) then the study add-ons are hidden. I do see them for just a moment when the page loads but then they are hidden and there are no available studies, regardless of whether I have already joined a study.

rhelmer:
Closes #235.

Checklist for reviewer:

### DIFF
--- a/src/routes/current-studies/Content.svelte
+++ b/src/routes/current-studies/Content.svelte
@@ -52,7 +52,7 @@ function leaveStudy(studyID) {
     <hr />
     <div class="studies">
 
-    {#each studies as study, i (study.addon_id)}
+    {#each studies as study}
     <StudyCard 
         title={study.name}
         author={study.authors.name}
--- a/tests/core-addon/integration/core_addon.js
+++ b/tests/core-addon/integration/core_addon.js
@@ -93,12 +93,9 @@ describe("Core-Addon Onboarding", function () {
     await findAndAct(this.driver, By.id("rally-core_mozilla_org-browser-action"), e => e.click());
 
     // We expect the extension to load its options page in a new tab.
-    console.debug("debug1");
     await this.driver.wait(async () => {
-      console.debug("window handles:", (await this.driver.getAllWindowHandles()).length);
       return (await this.driver.getAllWindowHandles()).length >= 2;
     }, WAIT_FOR_PROPERTY);
-    console.debug("debug2");
 
     // Selenium is still focused on the old tab, so switch to the new window handle.
     let latestTab = (await this.driver.getAllWindowHandles()).length - 1;
@@ -108,11 +105,9 @@ describe("Core-Addon Onboarding", function () {
     // Switch back to web content context.
     await this.driver.setContext(firefox.Context.CONTENT);
 
-    console.debug("debug3");
     // Ensure that the study card for the base study is displayed.
     const baseStudySelector = By.xpath(`//span[text()="Ion Base Study"]`);
     await this.driver.findElement(baseStudySelector);
-    console.debug("debug4");
 
     // Begin study unenrollment cancel it.
     await findAndAct(this.driver, By.xpath(`//button[text()="Leave Mozilla Rally"]`), e => e.click());
--- a/tests/core-addon/integration/core_addon.js
+++ b/tests/core-addon/integration/core_addon.js
@@ -86,13 +86,33 @@ describe("Core-Addon Onboarding", function () {
     await findAndAct(this.driver, By.css(`[label="Add"]`), e => e.click());
     await findAndAct(this.driver, By.css(`[label="Okay, Got It"]`), e => e.click());
 
-    // FIXME close tab and click on icon, check that post-enrollment options page is shown.
-    // This will currently fail because there is a bug in the core-addon UI, where
-    // the options page will show no studies.
-    // See https://github.com/mozilla-rally/core-addon/issues/235
+    // Close options page tab.
+    await this.driver.close();
+    const originalTab = (await this.driver.getAllWindowHandles())[0];
+    await this.driver.switchTo().window(originalTab);
+    await findAndAct(this.driver, By.id("rally-core_mozilla_org-browser-action"), e => e.click());
+
+    // We expect the extension to load its options page in a new tab.
+    console.debug("debug1");
+    await this.driver.wait(async () => {
+      console.debug("window handles:", (await this.driver.getAllWindowHandles()).length);
+      return (await this.driver.getAllWindowHandles()).length >= 2;
+    }, WAIT_FOR_PROPERTY);
+    console.debug("debug2");
+
+    // Selenium is still focused on the old tab, so switch to the new window handle.
+    let latestTab = (await this.driver.getAllWindowHandles()).length - 1;
+    const newOptionsTab = (await this.driver.getAllWindowHandles())[latestTab];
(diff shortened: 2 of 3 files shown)


src/routes/current-studies/Content.svelte
```diff
@@ -52,7 +52,7 @@ function leaveStudy(studyID) {
     <hr />
     <div class="studies">
 
-    {#each studies as study, i (study.addon_id)}
+    {#each studies as study}
```
rhelmer, line 55: @hamilton I'm not sure if this is the correct fix, but I could not figure out what the `{#each}` block here was doing based on the Svelte docs... we don't seem to be using the loop counter (`i`) and I'm not sure what `(study.addon_id)` does here?

rhelmer: Regardless of the fix, I think the test I've added will exercise this :)

hamilton: Yes, we don't need `i` but we will probably need the `(study.addon_id)` key. This tells us that if this value changes for this iterable element, to make sure to update the component. React has a similar concept in the `key` prop.

hamilton: Without this, if the study order changes (for instance once we make studies sortable, or if some other request changes the studies displayed through an async update) the rendered list of components won't update properly.

hamilton (changes requested):
Turns out this actually does not fix the problem. One of the consequences of the approach we're taking is that most bugs like this (I suspect) are largely going to be due to issues with the store + the background script not returning things in the right place. So the issue isn't actually in `Content.svelte`, but that's where we're seeing the error surface. 

But if we look upstream, we see an issue in `create-store.js`. The `api.getAvailableStudies`call appears to be returning the _whole_ initialized store, not just the studies paired w/ whether the user has joined any. I be the real issue is even more upstream from that. So the initialize call in that file should look like 

```javascript
  api.initialize(initialState).then(async (remoteInitialState) => {
    console.log('initialize', remoteInitialState, initialState);
    const state = remoteInitialState || initialState;
    const stateFromBg = await api.getAvailableStudies();
    if (stateFromBg.availableStudies) {
      state.availableStudies = stateFromBg.availableStudies;
    }    
    set(state);
  });
```

This seems to work but it seems like we should either get rid of `api.getAvailableStudies` or always ensure it returns just the available studies, and not the store.

The reason why it was failing in the Svelte template is because it was trying to iterate over a javascript object in `store.availableStudies` that was actually just a whole copy of the store, not the array of available studies we wanted to iterate over.

tests/core-addon/integration/core_addon.js
```diff
@@ -86,14 +86,34 @@ describe("Core-Addon Onboarding", function () {
...
+
+    // Selenium is still focused on the old tab, so switch to the new window handle.
+    let latestTab = (await this.driver.getAllWindowHandles()).length - 1;
+    const newOptionsTab = (await this.driver.getAllWindowHandles())[latestTab];
+    await this.driver.switchTo().window(newOptionsTab);
 
     // Switch back to web content context.
     await this.driver.setContext(firefox.Context.CONTENT);
 
+    console.debug("debug3");
```
Dexterp37, line 111: Stray debug log.

tests/core-addon/integration/core_addon.js
```diff
@@ -86,14 +86,34 @@ describe("Core-Addon Onboarding", function () {
...
-    // the options page will show no studies.
-    // See https://github.com/mozilla-rally/core-addon/issues/235
+    // Close options page tab.
+    await this.driver.close();
+    const originalTab = (await this.driver.getAllWindowHandles())[0];
+    await this.driver.switchTo().window(originalTab);
+    await findAndAct(this.driver, By.id("rally-core_mozilla_org-browser-action"), e => e.click());
+
+    // We expect the extension to load its options page in a new tab.
+    console.debug("debug1");
```
Dexterp37, line 96: Stray debug log.

tests/core-addon/integration/core_addon.js
```diff
@@ -86,14 +86,34 @@ describe("Core-Addon Onboarding", function () {
...
+    await this.driver.wait(async () => {
+      console.debug("window handles:", (await this.driver.getAllWindowHandles()).length);
+      return (await this.driver.getAllWindowHandles()).length >= 2;
+    }, WAIT_FOR_PROPERTY);
+    console.debug("debug2");
```
Dexterp37, line 101: Same.

tests/core-addon/integration/core_addon.js
```diff
@@ -86,14 +86,34 @@ describe("Core-Addon Onboarding", function () {
...
+    // Close options page tab.
+    await this.driver.close();
+    const originalTab = (await this.driver.getAllWindowHandles())[0];
+    await this.driver.switchTo().window(originalTab);
+    await findAndAct(this.driver, By.id("rally-core_mozilla_org-browser-action"), e => e.click());
+
+    // We expect the extension to load its options page in a new tab.
+    console.debug("debug1");
+    await this.driver.wait(async () => {
+      console.debug("window handles:", (await this.driver.getAllWindowHandles()).length);
```
Dexterp37, line 98: Here as well :)

tests/core-addon/integration/core_addon.js
```diff
@@ -86,14 +86,34 @@ describe("Core-Addon Onboarding", function () {
...
+    await this.driver.switchTo().window(newOptionsTab);
 
     // Switch back to web content context.
     await this.driver.setContext(firefox.Context.CONTENT);
 
+    console.debug("debug3");
+    // Ensure that the study card for the base study is displayed.
+    const baseStudySelector = By.xpath(`//span[text()="Ion Base Study"]`);
+    await this.driver.findElement(baseStudySelector);
+    console.debug("debug4");
```
Dexterp37, line 115: This, too.

Dexterp37 (changes requested):

rhelmer:
@hamilton thanks for the detailed info in https://github.com/mozilla-rally/core-addon/pull/280#pullrequestreview-557299906 !

I do see that the initial store is different from the updated store, and I feel like I have it pinpointed but I haven't quite figured it out yet.

The initial store looks like:
```
{
  enrolled: true,
  availableStudies: {
    enrolled: true,
    availableStudies: [...]
  }
}
```
which is... weird :slightly_smiling_face:

the updated state is like:
```
{
  enrolled: true,
  availableStudies: [...]
}
```
like you'd expect.

The initial data structure is what the code is expecting but it looks wrong, that's probably the real fix here. However @hamilton's sample patch in the comment above (while not ideal) works around it for now.

Since this is completely broken, I propose that we land the integration test and the workaround patch from above so we can do a dog-foodable release tomorrow, but leave the issue open and spend some more time on a proper fix.

@Dexterp37 wdyt?

Dexterp37:
> Since this is completely broken, I propose that we land the integration test and the workaround patch from above so we can do a dog-foodable release tomorrow, but leave the issue open and spend some more time on a proper fix.
> 
> @Dexterp37 wdyt?

I took a look and ended up removing some code along the way. We were misusing an API! I created #288. I would recommend closing this in favour of that one.

PR #283, opened 2021-01-05
use only needed woff2 font files for core add-on

### LINKED ISSUE #275: Do we need to package all the fonts in `public/fonts`?
When building the core addon XPI we currently package quite a few files in the `public/font` directory (20 fonts, ~1.59MB), accounting for ~91% of the add-on size. Do we need them all?

@hamilton ?

hamilton:
Closes #275. This brings down the size of the xpi from 1.8mb to about 550kb.

Checklist for reviewer:


Dexterp37 (approved):
r+ - thanks for doing this. Please make sure that the addon packs without new warnings and that the generated file runs without problems in the browser before merging. Thanks!

Issue #296, opened 2021-01-12
add support for RemoteSettings via privileged API


rhelmer:
Per https://docs.google.com/document/d/1hciaUOnMBRDrVyUaLf0Pjor8Zhtx0_BIoSL5dkMcgMQ we'd like to use the native Firefox `RemoteSettings` API to listen and respond to changes to a remote-settings collection (see issue #106).

This will allow us to update available add-ons from remote, as well as provide a "kill switch" (closes issue #132).

rhelmer:
In terms of API design, I am thinking of adding a listener, something like:

```
firefoxPrivilegedApi.remoteSettings.onMessage((data) => { ... });
```

On extension startup, the `RemoteSettings.on(...)` API will be used to listen for changes to the collection and fire an async callback with the latest data from the collection to all listeners. On extension shutdown, all listeners will be removed and `RemoteSettings.off(...)` will be used to stop listening for updates to the collection.

I don't anticipate we'll ever need more than one listener, but I think this is a better design than having the callback register a separate `RemoteSettings.on` listener for each and every listener.

From the Svelte side, we will update the list of collection in the store whenever the callback fires. Additionally, we will need a way to fetch the latest set of studies on startup:

```
const studies = await firefoxPrivilegedApi.remoteSettings.get();
```

Note that the collection is not passed as a parameter, and is hardcoded in the privileged API class. This is so that we do not have any mistakes, and also so that malicious code cannot get access to data outside of our collection (assuming the malicious code is only on the webextension side of the privilege boundary).

rhelmer:
I think this should cover normal install/uninstall and upgrade states, the only problems I can foresee is if the add-on is removed without uninstall being called (this can happen if the extension is removed directly from the filesystem for instance, which the user or anti-virus/malware).

Dexterp37:
Some feedback on the implementation (in hindsight, this should have probably been a chapter on the proposal!). I believe all of the RS implementation should live within the `core-addon/` part of Rally: Svelte should really know nothing about this.

[...]

Issue #132, opened 2020-11-04
Stop enrollment for users on specific versions of the core add-on


Dexterp37:
When we start using a core add-on to enroll study add-ons, there will likely be cases where we want to stop enrollment for users on some version of the core add-on because it has a bug and must be updated before we can proceed (either legal or technical).

For instance if we need to update our program consent dialog before we can enable certain partner studies. other than that just garden-variety technical bugs.

A technical bug would be like, we're going to accidentally through away data if we let users enroll version z study if the user has core add-on x installed so we need to update to core add-on y before unlocking it, or something like that.

This can be done either in the partner support library or remote settings (depending on the desired latency of days vs minutes).

marniepw:
Have kill switch ability via RS. Does this work with core add-on?

But if need to update study, operational processes here.

Need proposal, can be done here in ticket, but need to figure out ASAP.

rhelmer:
I've [started a proposal](https://docs.google.com/document/d/1hciaUOnMBRDrVyUaLf0Pjor8Zhtx0_BIoSL5dkMcgMQ/edit#) that covers this use case.

rhelmer:
Now that PR #311 has landed we should be able to implement this.

Per https://github.com/mozilla-rally/rally-core-addon/issues/132#issue-736305535 we need to be able to handle pausing and resuming studies for specific version(s) of the core add-on.

I'm thinking of handling this by adding a new attribute to the study metadata, something like:

```js
studyPaused: {
  "a.b.c": true, 
  "x.y.z": false,
}
```

Which would cause the study to be paused on version "a.b.c" but not version "x.y.z" of the core add-on. We could also allow `"*"` which would pause is for all versions of the core add-on.

Technically I am thinking that when a study is "paused" this means:

- it is hidden from the UI (no further enrollments allowed)
- any already-installed study add-ons are disabled

Assuming this all makes sense, the one bit I am unsure about is how disabling study add-ons should be implemented. It feels like the choices here to either:

1. disable the study extension using the browser API (unfortunately the [WebExtension API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/management/setEnabled) only works for themes not other extensions)
2. pass a message and have `rally.js` handle it in the study itself

I think given the lack of a standard API for (1), (2) is probably the way we should go. I think that means exactly what "disabled" means is ambiguous - presumably the study add-on needs to continue to listen for messages but doesn't need to do much else.

[...]

Issue #253, opened 2020-12-12
should the integration test pull from AMO?


rhelmer:
There is a certain value to testing that it works for real from AMO installs, but it is just another thing we need to keep track of, and adds latency another point of failure (network, site down, etc). Additionally, builds are not triggered when the study add-on changes.

@Dexterp37 and I discussed this in PR #228, the only two options we thought of at the time where to either:

1. build the study add-on repo
2. land a signed study add-on in the ion-core-addon repo

The problem with (1) is that it will be unsigned, so we will not be able to test that the prompt behavior is correct. (2) requires us to have a process for taking the signed study add-on from AMO and landing it in the repo, and is similar to what we do in Firefox to test this sort of behavior.

We decided not to tackle this yet, but it's something we should start thinking about.

Dexterp37:
I think we have another option here: we could generate the study addon just for tests, as part of the test preparation. We do this quite often in m-c to test features. I quickly experimented with the approach [here](https://github.com/mozilla-ion/ion-core-addon/compare/master...Dexterp37:local_study_test).

There's a few benefits:

- it would allow us to do a full end-to-end test (rally.js to core addon) always using the most recent version of the code, meaning that any wrong change would break the tests;
- we could use this approach just for some tests, having a special test to exclusively test the prompt-less experience with the signed addon (that would be fetched from AMO);

@rhelmer , thoughts?

marniepw:
Want to **not** pull from AMO for test, want to build the study template as part of the test.

PR #311, opened 2021-01-23
Use Remote Settings for fetching studies.

rhelmer:
Fixes issue #296, prereq for issue #132

Checklist for reviewer:

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,38 @@ To test the QA-signed extension in Firefox, you must be running the Nightly rele
 
 > **Note:** this will cause production-signed extensions (such as those from addons.mozilla.org) to not load. To allow these, set `xpinstall.signatures.required` pref to `false`.
 
+## Testing Remote Settings changes
+
+The Rally core-addon uses the Firefox [remote settings server](https://remote-settings.readthedocs.io/en/latest/) to publish metadata about current studies.
+Changes may be tested locally, or on the dev and staging servers, before pushing live.
+
+> **Note:** `npm run build-local-addon` will produce a build that uses `public/locally-available-studies.json` instead of fetching from Remote Settings. If you are doing local development you most likely want this option.
+
+For developer convenience, a [public remote-settings dev server](https://remote-settings.readthedocs.io/en/latest/tutorial-dev-server.html) is available. Making
+changes to staging or production require multiple sign-offs from Mozilla.
+
+You may also set up your own [local remote settings server](https://remote-settings.readthedocs.io/en/latest/tutorial-local-server.html).
+
+> **Note:** Firefox Release cannot be reconfigured to use a different server, an unbranded build such as Firefox Nightly must be used for testing.
+
+1. In `about:config` on Firefox Nightly, change `services.settings.server` to the URL for your server. If you're using the public remote-settings dev server
+described above, then the value will be `https://kinto.dev.mozaws.net/v1`.
+2. In the Firefox Browser console, first import the `RemoteSettings` module:
+```js
+const { RemoteSettings } = ChromeUtils.import(
+  "resource://services-settings/remote-settings.js"
+);
+```
+3. Then, disable signature checking and fetch the latest `rally-studies-v1` collection:
+```js
+RemoteSettings("rally-studies-v1").verifySignature = false;
+await RemoteSettings("rally-studies-v1").get();
+```
+4. The UI for the core add-on options page should respond immediately. After making changes to the RS server, you can either wait or explicitly poll for updates:
+```js
+await RemoteSettings.pollChanges()
+```
+
 ## Developing new frontend components
 
 The Core Add-On uses Storybook to assist in isolated component work. If you're building a new component, look at the examples in `/stories`. To run the storybook, run `npm run storybook`.
--- a/core-addon/Core.js
+++ b/core-addon/Core.js
@@ -11,29 +11,42 @@ const OPTIONS_PAGE_PATH = "public/index.html";
 const DEFAULT_ARGS = {
   // NOTE: if this URL ever changes, you will have to update the domain in
   // the permissions in manifest.json.
-  availableStudiesURI: "https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/pioneer-study-addons-v1/records",
+  availableStudiesURI: "https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/rally-studies-v1/records",
+  disableRemoteSettings: false,
   website: "https://mozilla-rally.github.io",
 }
 
 module.exports = class Core {
- /**
-  * @param {Object} args arguments passed in from the user.
-  * @param {String} args.availableStudiesURI the URI where the available studies
-  *             information is listed.
-  * @param {String} args.website the URL of the platform website.
-  */
+  /**
+   * @param {Object} args arguments passed in from the user.
+   * @param {String} args.availableStudiesURI the URI where the available studies
+   *             information is listed. Only used when disableRemoteSettings is `true`.
+   * @param {boolean} args.disableRemoteSettings do not use the official RemoteSettings server.
+   *             Default is `true`.
+   * @param {String} args.website the URL of the platform website.
+   */
   constructor(args = {}) {
-    this._userArguments = {...DEFAULT_ARGS, ...args};
+    this._userArguments = { ...DEFAULT_ARGS, ...args };
 
     this._storage = new Storage();
     this._dataCollection = new DataCollection();
 
     // Asynchronously get the available studies. We don't need to wait
     // for this to finish, the UI can handle the wait.
-    this._availableStudies =
-      this._fetchAvailableStudies()
-          .then(studies => this._updateInstalledStudies(studies));
-
+    this._availableStudies = this._fetchAvailableStudies().then((studies) =>
+      this._updateInstalledStudies(studies)
+    );
+    if (this._userArguments.disableRemoteSettings) {
+      console.warn("RemoteSettings disabled, not adding a new studies listener.");
+    } else {
+      // Register a listener to react to remote-settings updates.
+      browser.firefoxPrivilegedApi.onRemoteSettingsSync.addListener((studies) => {
+        // FIXME Important: this may be racing with the initial update or if there's two consecutive updates?
+        // @see https://github.com/mozilla-rally/rally-core-addon/issues/318
+        this._availableStudies = this._updateInstalledStudies(studies);
+        this._sendStateUpdateToUI();
+      });
+    }
     this._connectionPort = null;
   }
 
@@ -118,11 +131,11 @@ module.exports = class Core {
     // Update the available studies list with the installation
     // information.
     this._availableStudies = Promise.resolve(knownStudies.map(s => {
-        if (s.addon_id == info.id) {
-          s.studyInstalled = installed;
-        }
-        return s;
-      })
+      if (s.addon_id == info.id) {
+        s.studyInstalled = installed;
+      }
+      return s;
+    })
     );
 
     if (installed) {
@@ -179,7 +192,7 @@ module.exports = class Core {
     switch (message.type) {
       case "enrollment":
         return this._enroll()
-                   .then(r => this._sendStateUpdateToUI());
+          .then(r => this._sendStateUpdateToUI());
       case "get-studies":
         return this._sendStateUpdateToUI();
       case "study-unenrollment":
@@ -234,7 +247,7 @@ module.exports = class Core {
         };
       }
       case "telemetry-ping": {
-        const {payloadType, payload, namespace, keyId, key} = message.data;
+        const { payloadType, payload, namespace, keyId, key } = message.data;
         let rallyId = await this._storage.getRallyID();
         return await this._dataCollection.sendPing(
           rallyId, payloadType, payload, namespace, keyId, key
@@ -422,7 +435,7 @@ module.exports = class Core {
     await this._storage.clearActivatedStudies();
 
     // Finally, uninstall the addon.
-    await browser.management.uninstallSelf({showConfirmDialog: false});
+    await browser.management.uninstallSelf({ showConfirmDialog: false });
   }
 
   /**
@@ -473,7 +486,7 @@ module.exports = class Core {
    *          has at least the `addon_id` and `studyInstalled` properties.
    */
   async _updateInstalledStudies(studies) {
-    console.debug("Core._updateInstalledStudies");
+    console.debug("Core._updateInstalledStudies:", studies);
 
     // If we were able to fetch studies definitions, see if any
     // of them were installed. Start by getting the list of installed
@@ -481,7 +494,7 @@ module.exports = class Core {
     let installedAddonsIds =
       await browser.management.getAll().then(addons =>
         addons.filter(a => a.type == "extension")
-              .map(a => a.id));
+          .map(a => a.id));
     return studies.map(s => {
       s.studyInstalled = installedAddonsIds.includes(s.addon_id);
       return s;
@@ -492,14 +505,23 @@ module.exports = class Core {
    * Fetch the available studies.
    *
    * This loads the studies from the Firefox Remote Settings service.
+   * If `disableRemoteSettings` is `true`, then an URL will be used instead. This is intended for local testing.
    *
    * @returns {Promise(Array<Object>)} resolved with an array of studies
    *          objects, or an empty array on failures.
    */
   async _fetchAvailableStudies() {
     try {
-      const request = await fetch(this._userArguments.availableStudiesURI);
-      return (await request.json()).data;
+      let studies = [];
+      if (this._userArguments.disableRemoteSettings) {
+        console.warn("Not using RemoteSettings, fetching from:", this._userArguments.availableStudiesURI);
+        const request = await fetch(this._userArguments.availableStudiesURI);
+        studies = await request.json();
+      } else {
+        console.debug("Using RemoteSettings for studies.");
+        studies = await browser.firefoxPrivilegedApi.getRemoteSettings();
+      }
+      return studies;
     } catch (err) {
       console.error(err);
       return [];
@@ -556,7 +578,7 @@ module.exports = class Core {
 
     // Send a message to the UI to update the list of studies.
     this._connectionPort.postMessage(
-      {type: "update-state", data: newState});
+      { type: "update-state", data: newState });
   }
 
   /**
--- a/core-addon/FirefoxPrivilegedApi.js
+++ b/core-addon/FirefoxPrivilegedApi.js
@@ -4,16 +4,29 @@
 
 "use strict";
 
-const { TelemetryController } = ChromeUtils.import(
-  "resource://gre/modules/TelemetryController.jsm"
-);
 const { XPCOMUtils } = ChromeUtils.import(
   "resource://gre/modules/XPCOMUtils.jsm"
 );
 
+/* global RemoteSettings TelemetryController */
+XPCOMUtils.defineLazyModuleGetters(this, {
+  RemoteSettings: "resource://services-settings/remote-settings.js",
+  TelemetryController: "resource://gre/modules/TelemetryController.jsm",
+});
+
+/* global gUUIDGenerator */
 XPCOMUtils.defineLazyServiceGetters(this, {
   gUUIDGenerator: ["@mozilla.org/uuid-generator;1", "nsIUUIDGenerator"],
 });
+
+/**
+ * This is the Remote Settings collection key also known as "CID" in the official documentation.
+ * @see https://remote-settings.readthedocs.io/en/latest/getting-started.html
+ *
+ * NOTE - we cannot control which Remote Settings server that Firefox uses, so we assume that this
+ * is a global key that will be correct no matter which server is used (dev, staging, release, etc.)
+ */
+const STUDY_COLLECTION_KEY = "rally-studies-v1";
 
 this.firefoxPrivilegedApi = class extends ExtensionAPI {
   getAPI(context) {
@@ -29,11 +42,37 @@ this.firefoxPrivilegedApi = class extends ExtensionAPI {
           TelemetryController.submitExternalPing(type, payload, options);
         },
         async generateUUID() {
-          // eslint-disable-next-line no-undef
           let str = gUUIDGenerator.generateUUID().toString();
           return str.substring(1, str.length - 1);
-        }
-      }
-    }
+        },
+        async getRemoteSettings() {
+          return RemoteSettings(STUDY_COLLECTION_KEY).get();
+        },
+        // eslint-disable-next-line no-undef
+        onRemoteSettingsSync: new ExtensionCommon.EventManager({
+          context,
+          name: "firefoxPrivilegedApi.onRemoteSettingsSync",
+          // Fire a callback to listener when remote settings collection updates, passing the contents of the collection.
+          register: (fire) => {
+            const callback = (current) => {
+              fire.async(current).catch(() => { }); // ignore Message Manager disconnects
+            };
+            async function rsCallback(event) {
+              let {
+                data: { current },
+              } = event;
+              callback(current);
+            }
+            RemoteSettings(STUDY_COLLECTION_KEY).on("sync", rsCallback);
+            return () => {
+              RemoteSettings(STUDY_COLLECTION_KEY).off("sync", rsCallback);
+            };
+          },
+        }).api(),
+      },
+    };
+  }
+  onShutdown() {
+    RemoteSettings(STUDY_COLLECTION_KEY).db.clear()
   }
 };
--- a/core-addon/FirefoxPrivilegedApi.schema.json
+++ b/core-addon/FirefoxPrivilegedApi.schema.json
@@ -2,6 +2,13 @@
   {
     "namespace": "firefoxPrivilegedApi",
     "description": "experimental API extensions for exposing privileged Firefox APIs",
+    "events": [
+      {
+        "name": "onRemoteSettingsSync",
+        "type": "function",
+        "parameters": []
+      }
+    ],
     "functions": [
       {
         "name": "submitEncryptedPing",
@@ -63,6 +70,13 @@
         "name": "generateUUID",
         "type": "function",
         "description": "Generate a random UUID using the Firefox APIs",
+        "parameters": [],
+        "async": true
+      },
+      {
+        "name": "getRemoteSettings",
+        "type": "function",
+        "description": "Get a RemoteSettings collection",
         "parameters": [],
         "async": true
       }
--- a/package.json
+++ b/package.json
@@ -2,9 +2,9 @@
   "name": "rally_core",
   "version": "0.8.0",
   "scripts": {
-    "build": "rollup -c && rollup -c rollup.config.addon.js --config-study-list-url=/public/locally-available-studies.json && web-ext build --overwrite-dest && mv web-ext-artifacts/*.zip web-ext-artifacts/rally_core.xpi",
+    "build": "rollup -c && npm run build-addon && web-ext build --overwrite-dest && mv web-ext-artifacts/*.zip web-ext-artifacts/rally_core.xpi",
     "build-addon": "rollup -c rollup.config.addon.js",
-    "build-local-addon": "rollup -c rollup.config.addon.js --config-study-list-url=/public/locally-available-studies.json",
+    "build-local-addon": "rollup -c rollup.config.addon.js --config-disable-remote-settings --config-study-list-url=/public/locally-available-studies.json",
     "build-storybook": "build-storybook -s ./",
     "lint": "npm run build && npm-run-all lint-*",
     "lint-addon": "web-ext lint",
@@ -15,7 +15,7 @@
     "storybook": "start-storybook -p 6006 -s ./",
     "test": "npm run --silent skip-taskcluster || (npm run test-addon && npm run test-support-library)",
     "test-addon": "npm run --silent skip-taskcluster || (mocha --require  \"./tests/hooks.js\" \"./tests/core-addon/unit/*.js\")",
-    "test-integration": "npm run --silent skip-taskcluster || (npm run build && mocha --timeout 30000 \"./tests/core-addon/integration/*.js\")",
+    "test-integration": "npm run --silent skip-taskcluster || (npm run build && mocha --unhandled-rejections=strict --timeout 30000 \"./tests/core-addon/integration/*.js\")",
     "test-support-library": "npm run --silent skip-taskcluster || (mocha --require  \"./tests/hooks.js\" \"./tests/support/*.mjs\")"
   },
   "devDependencies": {
--- a/public/locally-available-studies.json
+++ b/public/locally-available-studies.json
@@ -1,37 +1,35 @@
-{
-  "data": [
-    {
-      "name": "Rally Base Study",
-      "icons": {
-        "32": "https://addons.cdn.mozilla.net/static/img/addon-icons/default-32.png",
-        "64": "https://addons.cdn.mozilla.net/static/img/addon-icons/default-64.png",
-        "128": "https://addons.cdn.mozilla.net/static/img/addon-icons/default-128.png"
-      },
-      "schema": 1597266497978,
-      "authors": {
-        "url": "https://addons.mozilla.org/en-US/firefox/addon/ion-basic-study/",
-        "name": "Ion Developers"
-      },
-      "version": "0.3.1",
-      "addon_id": "ion-basic-study@mozilla.org",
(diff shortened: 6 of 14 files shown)


rhelmer:
@Dexterp37 OK so this works (well outside of the listener part but I am confident we can figure that out).

However, looking at this now it occurs to me that we could do this in unprivileged code by just fetching the appropriate JSON directly from the Remote Settings CDN and implement the listener to fire on a timer that only runs when the user is idle.

The overhead of using the built-in Firefox API doesn't seem worth it, the only advantage is really that it supports push but that doesn't really seem very compelling (some clients can't use push and fall back to an RS timer, anyway).

@Dexterp37 wdyt?

rhelmer:
> @Dexterp37 OK so this works (well outside of the listener part but I am confident we can figure that out).
> 
> However, looking at this now it occurs to me that we could do this in unprivileged code by just fetching the appropriate JSON directly from the Remote Settings CDN and implement the listener to fire on a timer that only runs when the user is idle.

We may need to use https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/alarms if we go this route. It might mean another permission, but there are some restrictions on the web content APIs per that doc.

core-addon/Core.js
```diff
@@ -33,7 +33,10 @@ module.exports = class Core {
     this._availableStudies =
       this._fetchAvailableStudies()
           .then(studies => this._updateInstalledStudies(studies));
-
+            // Register a listener to react to remote-settings updates.
+            browser.firefoxPrivilegedApi.onRemoteSettingsSync.addListener((result) => {
+              this._sendStateUpdateToUI(result[0].data);
+          });
```
Dexterp37, line 39: nit: the indentation looks odd here :-)

core-addon/FirefoxPrivilegedApi.js
```diff
@@ -4,17 +4,24 @@
...
 );
 
+const { RemoteSettings } = ChromeUtils.import(
+  "resource://services-settings/remote-settings.js"
+);
+
+const { TelemetryController } = ChromeUtils.import(
+  "resource://gre/modules/TelemetryController.jsm"
+);
+
```
Dexterp37, line 18: Why was this moved? (Was linting complaining or something?) - Nothing to do, but I'm curious :)

core-addon/FirefoxPrivilegedApi.js
```diff
@@ -4,17 +4,24 @@
...
 XPCOMUtils.defineLazyServiceGetters(this, {
   gUUIDGenerator: ["@mozilla.org/uuid-generator;1", "nsIUUIDGenerator"],
 });
 
+const STUDY_COLLECTION_KEY = "rally-studies-v1";
```
Dexterp37, line 23: nit: can we add a one line comment explaining what's this for?

core-addon/FirefoxPrivilegedApi.schema.json
```diff
@@ -2,6 +2,13 @@
   {
     "namespace": "firefoxPrivilegedApi",
     "description": "experimental API extensions for exposing privileged Firefox APIs",
+    "events": [
```
Dexterp37, line 5: ah! TIL! I'm not an expert of this privilege API exposure. Let me bring in somebody else :)

core-addon/FirefoxPrivilegedApi.js
```diff
@@ -32,8 +39,30 @@ this.firefoxPrivilegedApi = class extends ExtensionAPI {
           // eslint-disable-next-line no-undef
           let str = gUUIDGenerator.generateUUID().toString();
           return str.substring(1, str.length - 1);
-        }
-      }
-    }
+        },
+        async getRemoteSettings() {
+          return RemoteSettings(STUDY_COLLECTION_KEY).get();
```
Dexterp37, line 44: Can `RemoteSettings(STUDY_COLLECTION_KEY)` ever fail/return `undefined` or `null`?

core-addon/Core.js
```diff
@@ -498,8 +501,9 @@ module.exports = class Core {
    */
   async _fetchAvailableStudies() {
     try {
-      const request = await fetch(this._userArguments.availableStudiesURI);
-      return (await request.json()).data;
+      const result = await browser.firefoxPrivilegedApi.getRemoteSettings();
+      const studies = result[0].data;
```
Dexterp37, line 505: Do we need some defensive programming here? Is `result` guaranteed to be a valid array that has at least an item with the `data` property?

Dexterp37 (commented):
This approach looks good to me! I'd just love other experts to validate:

- our usage of the `RemoteSettings`
- the exposed privileged API (someone that works with addons API, maybe :rpl/Luca Greco)

> 
> 
> @Dexterp37 OK so this works (well outside of the listener part but I am confident we can figure that out).

What's wrong with the listener?

> However, looking at this now it occurs to me that we could do this in unprivileged code by just fetching the appropriate JSON directly from the Remote Settings CDN and implement the listener to fire on a timer that only runs when the user is idle.
> 
> The overhead of using the built-in Firefox API doesn't seem worth it, the only advantage is really that it supports push but that doesn't really seem very compelling (some clients can't use push and fall back to an RS timer, anyway).
> 
> @Dexterp37 wdyt?

I agree on this second point. However, there might be other edge cases/trickeries that the Firefox implementation subtly takes care of (for us, thankfully!). An unprivileged API is definitely a blocker for future expansions. What about we keep the current implementation, fix the tests/listeners and file a follow-up issue to re-evaluate once we _require_ this new step?

core-addon/FirefoxPrivilegedApi.schema.json
Dexterp37: Hey @rpl, would you kindly review the changes to this "experimental API"?

core-addon/FirefoxPrivilegedApi.js
rhelmer: Just putting them in lexical order :) I wonder if there's an eslint rule for that.

core-addon/Core.js
rhelmer: Hm yes my vscode eslint seems to disagree with ours an I haven't figured out why yet, so I have not been using auto-format :( Will figure this out.

rhelmer:
> This approach looks good to me! I'd just love other experts to validate:
> 
>     * our usage of the `RemoteSettings`
> 
>     * the exposed privileged API (someone that works with addons API, maybe :rpl/Luca Greco)
> 
> 
> > @Dexterp37 OK so this works (well outside of the listener part but I am confident we can figure that out).
> 
> What's wrong with the listener?

Sorry, the listener itself is fine and fires when I run `RemoteSettings.pollChanges()` as expected, but the UI does not update.

> > However, looking at this now it occurs to me that we could do this in unprivileged code by just fetching the appropriate JSON directly from the Remote Settings CDN and implement the listener to fire on a timer that only runs when the user is idle.
> > The overhead of using the built-in Firefox API doesn't seem worth it, the only advantage is really that it supports push but that doesn't really seem very compelling (some clients can't use push and fall back to an RS timer, anyway).
> > @Dexterp37 wdyt?
> 
> I agree on this second point. However, there might be other edge cases/trickeries that the Firefox implementation subtly takes care of (for us, thankfully!). An unprivileged API is definitely a blocker for future expansions. What about we keep the current implementation, fix the tests/listeners and file a follow-up issue to re-evaluate once we _require_ this new step?

WFM! The good news is that it looks pretty straightforward either way for the implementation, test coverage is what I am figuring out now.

I guess one concern I have is if we're leaving some kind of leftover trace in the Firefox `RemoteSettings` internal storage, I'll ask @leplatrem about that part specifically.

core-addon/Core.js
rhelmer: Yes, this is definitely a "WIP" bit :) I'd like to unpack the datastructure in a more sane way. I may move this to the privileged API implementation, so it returns something that's more useful to us.

For `about:ion` we only care about maintaining a single collection (not a set of collections in a record which is quite common), so we could do something much simpler here.

core-addon/FirefoxPrivilegedApi.js
rhelmer: It should always return `[]` I believe, if there's any problem, but I'll double-check this.

core-addon/FirefoxPrivilegedApi.schema.json
rhelmer: Thanks :) FTR the `EventManager` thing is great :)

core-addon/Core.js
rhelmer: Oh @Dexterp37 so as I mentioned in a separate comment, I'm pretty sure I'm passing the right data here (the same as `_fetchAvailableStudies()` returns) but I don't see the UI update, I have been focusing on getting all the RS bits right but I haven't figured this out yet.

core-addon/FirefoxPrivilegedApi.js
rhelmer: I guess by that logic `XPCOMUtils` should be moved below but it's a common practice to have it be the first so it can be used for lazy imports (which I will likely switch these to)

rhelmer:
> I guess one concern I have is if we're leaving some kind of leftover trace in the Firefox `RemoteSettings` internal storage, I'll ask @leplatrem about that part specifically.

After going for a walk and giving this more though :) so it's not usually possible for webextensions to reliably run code on shutdown, we should be able to do so for privileged extensions. So if there's any clean up to do we can do it there.

However! There is one edge case which we may or may not want to spend a lot of time on:

1. Rally core add-on is installed
2. Firefox is shut down
3. Rally add-on `.xpi` is removed from the filesystem
4. Firefox starts up and registers that the extension is not installed

Since the code is removed entirely in (3) Firefox is not able to run any sort of shutdown code in (4). If we're using pure webextension APIs and storage and such Firefox handles this for us in (4) but the privileged code is simply gone so there isn't a great way to do that.

In any case it might not be an issue here (it's possible there's no cleanup needed for the internal RS) but this is something we should verify.

rhelmer:
> I agree on this second point. However, there might be other edge cases/trickeries that the Firefox implementation subtly takes care of (for us, thankfully!). An unprivileged API is definitely a blocker for future expansions. What about we keep the current implementation, fix the tests/listeners and file a follow-up issue to re-evaluate once we _require_ this new step?

Oh shoot actually you know what, I missed a very important detail here which is content signing. The built-in Firefox RS API ships with a key and does verification of the content this way, which we also want (otherwise a MITM could serve a false list of studies as official).

So yes the pure-content is not as simple as it appeared :) We should consider making a library for this or reviving one of the current ones, but let's not rabbit-hole on this and move forward with the privileged approach as you've suggested.

core-addon/FirefoxPrivilegedApi.js
rhelmer: I went a little overboard but I think it's probably appropriate :)

core-addon/Core.js
rhelmer: We could consider something like having a separate record for each study, but I'm not sure that's helpful since we really want to be able to ensure that a given set of studies are interoperable, it just seems like more overhead and work to have to do.

It's essentially the difference between maintaining one single large JSON file vs. a bunch of small independent JSON files, the latter feels a little cleaner somehow but I think it'd just be overhead with no upside in our case.

core-addon/FirefoxPrivilegedApi.js
Dexterp37: Makes sense!

core-addon/Core.js
Dexterp37: > It's essentially the difference between maintaining one single large JSON file vs. a bunch of small independent JSON files, the latter feels a little cleaner somehow but I think it'd just be overhead with no upside in our case.

I agree!

core-addon/Core.js
```diff
@@ -30,13 +30,13 @@ module.exports = class Core {
...
-            browser.firefoxPrivilegedApi.onRemoteSettingsSync.addListener((result) => {
-              this._sendStateUpdateToUI(result[0].data);
-          });
+    this._availableStudies = this._fetchAvailableStudies().then((studies) =>
+      this._updateInstalledStudies(studies)
+    );
+    // Register a listener to react to remote-settings updates.
+    browser.firefoxPrivilegedApi.onRemoteSettingsSync.addListener((result) => {
+      this._sendStateUpdateToUI(result[0].data);
+    });
```
Dexterp37, line 39: This is not working because `_sendStateUpdateToUI` doesn't take an argument. To make responding to changes work, you actually need to do something like this:

```suggestion
    browser.firefoxPrivilegedApi.onRemoteSettingsSync.addListener(async function (result) {
            // Important: this may be racing with the initial update or if there's two consecutive updates?
            this._availableStudies = await this._updateInstalledStudies(result[0].data);
            this._sendStateUpdateToUI();
    });
```

Dexterp37 (changes requested):

core-addon/FirefoxPrivilegedApi.js
leplatrem: There is not global `try/finally` on the method, but yes, the returned value should always be a list ([source](https://searchfox.org/mozilla-central/rev/69babd862de70cabfa1d0a369d38e4881bd41e4d/services/settings/RemoteSettingsClient.jsm#343))

core-addon/FirefoxPrivilegedApi.js
```diff
@@ -26,14 +39,41 @@ this.firefoxPrivilegedApi = class extends ExtensionAPI {
...
+            // eslint-disable-next-line no-undef
+            RemoteSettings(STUDY_COLLECTION_KEY).on("sync", async (event) => {
+              let {
+                data: { current },
+              } = event;
+              callback(current);
+            });
+            return () => {
+              // eslint-disable-next-line no-undef
+              RemoteSettings(STUDY_COLLECTION_KEY).off("sync");
```
leplatrem, line 72: I just realize that `.off()` expects a callback reference :( https://searchfox.org/mozilla-central/rev/69babd862de70cabfa1d0a369d38e4881bd41e4d/services/settings/RemoteSettingsClient.jsm#114

leplatrem (changes requested):
> I guess one concern I have is if we're leaving some kind of leftover trace in the Firefox RemoteSettings internal storage, I'll ask @leplatrem about that part specifically.

Leaving traces in Chrome IDB is not really an issue. Data is indexed by collection name. The only impact is that the client will continue to sync the data from the server.

If that's an issue, the collection data can be cleared with  `await RemoteSettings(STUDY_COLLECTION_KEY).db.clear()`

core-addon/Core.js
rhelmer: aha, thank you! I figured I was just doing something wrong.

rhelmer: (well I can't take this suggestion as-is since this is a constructor so we can't `await` but I understand what you mean :) )

rhelmer: Ugh OK well you are using an `async` function for the callback, nevermind! `await` sounds great.

rhelmer:
Hrm OK so one thing I didn't anticipate that touches existing code is that https://github.com/mozilla-extensions/rally-core-addon/blob/master/public/locally-available-studies.json is what we've been testing against and is the direct result of `fetch()`ing from the RS server, but I think we want to make things a little nicer and have `firefoxPrivilegedApi.getRemoteSettings()` and the listener just pass an array. This is why we had to unpack the data structure more than it seemed like was necessary.

I'll clean this up now in a separate commit, I think it's only going to get more confusing if we try to work around it.

rpl (commented):
@Dexterp37 follows a couple of comments, I forwarded a more complete review pass on this to @mixedpuppy (especially because he did review experimental apis more often than me in the past and he may spot things that I may miss)

core-addon/FirefoxPrivilegedApi.js
```diff
@@ -4,17 +4,30 @@
...
 
-const { TelemetryController } = ChromeUtils.import(
-  "resource://gre/modules/TelemetryController.jsm"
-);
 const { XPCOMUtils } = ChromeUtils.import(
   "resource://gre/modules/XPCOMUtils.jsm"
 );
 
+// FIXME we shouldn't be failing no-undef eslint here, maybe we are not
+//       importing all the Firefox eslint rules properly?
```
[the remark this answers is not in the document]
rhelmer: Ah yeah looks like we're just going the `globals` approach:
https://github.com/mozilla-rally/core-addon/blob/master/.eslintrc.js#L17

I'll just do that here then thanks :)

core-addon/Core.js
rhelmer: I may have mentioned this elsewhere, but I figured this out and it was having an incorrect document that we've been testing against (it includes the "envelope" from RS which we probably don't want, but that's what you get when you pull directly from the server via `fetch()`)

We should have a JSON schema and check against it IMHO, both for development and also for production.

core-addon/FirefoxPrivilegedApi.js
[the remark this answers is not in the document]
rhelmer: Hm so that's a little frustrating, something like:
```
const { XPCOMUtils } = ChromeUtils.import(
  "resource://gre/modules/XPCOMUtils.jsm"
);
```

works OK because it's defining that new `const` so eslint is cool with it (since `ChromeUtils` is allowed as a global) but of course doing it this way:

```
XPCOMUtils.defineLazyModuleGetters(this, {
  RemoteSettings: "resource://services-settings/remote-settings.js",
  TelemetryController: "resource://gre/modules/TelemetryController.jsm",
});
```

Means that `RemoteSettings` and `TelemetryController` and anything else we import would need to be added to the allowed globals, which seems... not great :) So it might make more sense to just use comments, so we don't bloat the eslint config.

core-addon/Core.js
rhelmer: As I was implementing this, I had to refresh my memory about how we use RS in `about:ion`:

1. we *are* using a set of records in the collection, one for each study
2. the `RemoteSettings(KEY).get()` API returns all records together as an `Array` in JS

I reviewed the production config and set up the dev server correctly, and the current changeset seems to be working, including updating when RS does.

rhelmer:
This seems to be working overall (including updating real-time as soon as RS does!) except that while the core add-on does list all available studies in the UI, it not seem to trust the list of study add-ons anymore for some reason. I'll track that down in the morning.

Once that's figured out, I think all that's left are tests. There were a few things that I will split out to separate issues:

1. joining Rally fails if there are no available studies (if RS returns `[]`, which is expected upon error)
2. we should have a shared schema with the server, and check against it at test- and run-time.

rhelmer:
OK finally have a green build :) I've added unit and integration tests (and refactored the integration tests some), but I am cheating and using the RS dev server ;) I'm going to figure out how to do this locally, I don't think we need to run an actual RS server though just something that mocks the API well enough for integration tests is probably fine.

Dexterp37:
> 
> 
> OK finally have a green build :) I've added unit and integration tests (and refactored the integration tests some), but I am cheating and using the RS dev server ;) I'm going to figure out how to do this locally, I don't think we need to run an actual RS server though just something that mocks the API well enough for integration tests is probably fine.

Yeah, that's a good intuition. There's a couple of strategies used in m-c.

1. Directly add data to the RemoteSettings DB ([Normandy Example](https://searchfox.org/mozilla-central/rev/bd92b9b4a3c2ff022e830c1358968a84e6e69c95/toolkit/components/normandy/test/browser/browser_RecipeRunner.js#292-306), [Firefox Monitor Example](https://searchfox.org/mozilla-central/rev/bd92b9b4a3c2ff022e830c1358968a84e6e69c95/browser/components/fxmonitor/test/browser/browser_fxmonitor_doorhanger.js#57-69)).
2. Using [sinon JS to stub the return values](https://searchfox.org/mozilla-central/rev/bd92b9b4a3c2ff022e830c1358968a84e6e69c95/browser/components/extensions/test/xpcshell/test_ext_settings_overrides_defaults.js#131-142).

rhelmer:
> > OK finally have a green build :) I've added unit and integration tests (and refactored the integration tests some), but I am cheating and using the RS dev server ;) I'm going to figure out how to do this locally, I don't think we need to run an actual RS server though just something that mocks the API well enough for integration tests is probably fine.
> 
> Yeah, that's a good intuition. There's a couple of strategies used in m-c.
> 
>     1. Directly add data to the RemoteSettings DB ([Normandy Example](https://searchfox.org/mozilla-central/rev/bd92b9b4a3c2ff022e830c1358968a84e6e69c95/toolkit/components/normandy/test/browser/browser_RecipeRunner.js#292-306), [Firefox Monitor Example](https://searchfox.org/mozilla-central/rev/bd92b9b4a3c2ff022e830c1358968a84e6e69c95/browser/components/fxmonitor/test/browser/browser_fxmonitor_doorhanger.js#57-69)).
> 
>     2. Using [sinon JS to stub the return values](https://searchfox.org/mozilla-central/rev/bd92b9b4a3c2ff022e830c1358968a84e6e69c95/browser/components/extensions/test/xpcshell/test_ext_settings_overrides_defaults.js#131-142).
> 
> 
> I don't have any strong opinion on which technique to use. I'd say I'd slightly lean more towards 1, but we're in gut-feelings land.

Nice tip, thanks! I just used the `RemoteSettings`[...]`.db`method, it's pretty much a drop-in replacement for my "local server" method.

rhelmer:
@Dexterp37 I think this is pretty close, do you see any major issues? I feel like we should defer setting up a schema and simplifying the RS payload to a separate issue, in the interest of getting this landed soon. wdyt?

README.md
```diff
@@ -87,6 +87,34 @@ To test the QA-signed extension in Firefox, you must be running the Nightly rele
 
 > **Note:** this will cause production-signed extensions (such as those from addons.mozilla.org) to not load. To allow these, set `xpinstall.signatures.required` pref to `false`.
 
+## Testing Remote Settings changes
+
+The Rally core-addon uses the Firefox [remote settings server](https://remote-settings.readthedocs.io/en/latest/) to publish metadata about current studies.
+Changes may be tested locally, or on the dev and staging servers, before pushing live.
```
Dexterp37, line 93: What's needed for testing locally? Can any HTTP server work (by just serving the JSON file)?

core-addon/Core.js
```diff
@@ -30,10 +30,16 @@ module.exports = class Core {
...
-    this._availableStudies =
-      this._fetchAvailableStudies()
-          .then(studies => this._updateInstalledStudies(studies));
+    this._availableStudies = this._fetchAvailableStudies().then((studies) =>
+      this._updateInstalledStudies(studies)
+    );
 
+    // Register a listener to react to remote-settings updates.
+    browser.firefoxPrivilegedApi.onRemoteSettingsSync.addListener((studies) => {
+      // FIXME Important: this may be racing with the initial update or if there's two consecutive updates?
```
Dexterp37, line 39: Would you kindly file an issue about this and reference it here?

package.json
```diff
@@ -15,7 +15,7 @@
     "storybook": "start-storybook -p 6006 -s ./",
     "test": "npm run --silent skip-taskcluster || (npm run test-addon && npm run test-support-library)",
     "test-addon": "npm run --silent skip-taskcluster || (mocha --require  \"./tests/hooks.js\" \"./tests/core-addon/unit/*.js\")",
-    "test-integration": "npm run --silent skip-taskcluster || (npm run build && mocha --timeout 30000 \"./tests/core-addon/integration/*.js\")",
+    "test-integration": "npm run --silent skip-taskcluster || (npm run build && mocha --unhandled-rejections=strict --timeout 30000 \"./tests/core-addon/integration/*.js\")",
```
Dexterp37, line 18: Ah, nice. If I understand this correctly, tests should fail earlier with this flag in case of problems, without being falsely marked as positive. Is that correct? 

Would you kindly file an issue for adding this mode to the other test commands as well?

public/locally-available-studies.json
```diff
@@ -1,37 +1,35 @@
...
-      },
-      "joinStudyConsent": "This study will send an encrypted ping, only when the toolbar icon is clicked.",
-      "leaveStudyConsent": "This study cannot be re-joined.",
-      "dataCollectionDetails": ["The date and time"],
-      "id": "0eb02750-7159-4f09-96ae-5c7cb7424e89",
-      "last_modified": 1597280277565
-    }
-  ]
-}
+[
```
Dexterp37, line 1: Does this mean the `npm run build-local-addon` command would still work after these changes?

package.json
```diff
@@ -15,7 +15,7 @@
     "storybook": "start-storybook -p 6006 -s ./",
```
Dexterp37, line 15: Would you kindle file an issue about changing the `build` command to not use `--config-study-list-url` anymore? Our final releases should point to the remote settings server and the local list should exclusively be used for local testing.

We might want to keep this around for some more time to have an end-to-end QA loop.

public/locally-available-studies.json
```diff
@@ -1,37 +1,35 @@
...
+      "name": "Ion Developers"
+    },
+    "version": "0.3.1",
+    "addon_id": "ion-basic-study@mozilla.org",
+    "moreInfo": {
+      "spec": "https://addons.mozilla.org/en-US/firefox/addon/ion-basic-study/"
+    },
+    "isDefault": false,
+    "sourceURI": {
+      "spec": "https://addons.mozilla.org/firefox/downloads/file/3707495/ion_basic_study-0.3.1-fx.xpi"
```
Dexterp37, line 21: Since we're changing this file, what about changing this link to `https://addons.mozilla.org/firefox/downloads/latest/ion-basic-study` so that it always downloads the latest version? (this way we don't have to cut a new release of the core addon for template changes while we are in QA mode..)
```suggestion
      "spec": "https://addons.mozilla.org/firefox/downloads/latest/ion-basic-study"
```

tests/core-addon/integration/core_addon.js
```diff
@@ -28,63 +28,154 @@ async function findAndAct(driver, element, action) {
...
+  await this.driver.setContext(firefox.Context.CONTENT);
+}
 
-    // New tab is focused.
-    await this.driver.wait(
-      until.titleIs("Rally: Put your data to work for a better internet"),
-      WAIT_FOR_PROPERTY
-    );
+// A default set of test studies.
+// FIXME - this currently loads data from remote! We need to figure out how to host this locally.
```
Dexterp37, line 94: What data is loaded remotely? Do you mean the basic study?

tests/core-addon/integration/core_addon.js
```diff
@@ -28,63 +28,154 @@ async function findAndAct(driver, element, action) {
...
+//         `file://` URLs might work, but we may need to resort to using a local webserver.
+const rallyTestStudy = {
```
Dexterp37, line 96: Is there any value in re-using the json file we expose in `public/`?

Dexterp37 (commented):
This looks really good overall! I left a few comments below, they are all minor. I'm happy to take another look after they're addressed, I think we might merge shortly after that.

README.md
rhelmer: Hmm maybe... I was able to get it working for tests at least :) Honestly it's probably easiest to use the dev server I can link to https://remote-settings.readthedocs.io/en/latest/tutorial-dev-server.html and maybe https://remote-settings.readthedocs.io/en/latest/tutorial-local-server.html for setting up a local server.

package.json
rhelmer: Oh shoot, I forgot to deal cleaning up around the local list, thanks... I've given this some thought and actually I'm not sure if it will reduce or increase confusion having multiple ways to do this. I can make the local list continue working in the meantime while we figure this out.

package.json
rhelmer: Sure! Yes essentially it just errors out if any promises reject and it's never handled by anything, I noticed this when I forgot to do an `await` but the tests seemed to pass anyway (it was a legit error in the test),

public/locally-available-studies.json
rhelmer: Hm, the way this is currently implemented I think remote-settings is always used... per my comment above I will just go ahead and implement the old behavior in dev mode so it can still work, but we're going to need to be careful to build our production XPIs properly. Shouldn't be a big deal.

public/locally-available-studies.json
rhelmer: We can, the problem though is that the `version` field will then be wrong... that field isn't really used for anything right now but it's a consideration.

tests/core-addon/integration/core_addon.js
rhelmer: Yes, the local study and the icons. Everything with an `https://` link in here needs to be changed to either hit a local server, or we can see if `file://` works. I think it probably does.

tests/core-addon/integration/core_addon.js
rhelmer: Yes and I had intended to do so thanks :)

README.md
rhelmer: Although, as we discussed in other comment threads here, if we allow the `build-local-addon` feature to continue working then in general for folks building locally in dev mode it'd be easiest to use that method.

Using an actual RS server is probably more for people doing QA on builds, or testing something very specific to our RS integration.

package.json
rhelmer: issue #317

tests/core-addon/integration/core_addon.js
rhelmer: This is basically https://github.com/mozilla-rally/rally-core-addon/issues/253 which I think needs a followup issue?

rhelmer: https://github.com/mozilla-rally/rally-core-addon/issues/319

Dexterp37 (changes requested):
This looks good overall, there's a few things that need to be addressed and small nits here and there. I'll make sure to review this one last time and give the final r+ today!

core-addon/Core.js
```diff
@@ -11,30 +11,43 @@ const OPTIONS_PAGE_PATH = "public/index.html";
...
+   * @param {Object} args arguments passed in from the user.
+   * @param {String} args.availableStudiesURI the URI where the available studies
+   *             information is listed. Only used when enableRemoteSettings is `false`.
+   * @param {boolean} args.enableRemoteSettings use the privileged RemoteSettings server.
+   *             Default is `true`.
+   * @param {String} args.website the URL of the platform website.
+   */
   constructor(args = {}) {
-    this._userArguments = {...DEFAULT_ARGS, ...args};
+    this._userArguments = { ...DEFAULT_ARGS, ...args };
```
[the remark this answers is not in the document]
rhelmer: https://github.com/mozilla-rally/rally-core-addon/issues/327

core-addon/Core.js
```diff
@@ -179,7 +192,7 @@ module.exports = class Core {
     switch (message.type) {
       case "enrollment":
         return this._enroll()
-                   .then(r => this._sendStateUpdateToUI());
+          .then(r => this._sendStateUpdateToUI());
```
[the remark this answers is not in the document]
rhelmer: https://github.com/mozilla-rally/rally-core-addon/issues/327

tests/core-addon/integration/core_addon.js
```diff
@@ -28,63 +29,121 @@ async function findAndAct(driver, element, action) {
...
-  it("should un/enroll in Rally", async function () {
-    this.driver = await utils.getFirefoxDriver(true, {});
+/**
+ * Install the Rally extension, but do not join Rally.
+ *
+ * Leaves the Selenium driver context set to CONTENT, and
+ * focused on the options page.
+ */
+async function installRally(driver) {
+  this.driver = driver;
```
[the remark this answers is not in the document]
rhelmer: I did this initially so it'd be easier to refactor without a bunch of little changes, but yeah I can go ahead and do that.

tests/core-addon/integration/core_addon.js
```diff
@@ -28,63 +29,121 @@ async function findAndAct(driver, element, action) {
...
-    await this.driver.wait(async () => {
-      return (await this.driver.getAllWindowHandles()).length === 2;
-    }, WAIT_FOR_PROPERTY);
+  await findAndAct(this.driver, By.xpath(`//button[text()="Save & Continue"]`), e => e.click());
 
-    // Selenium is still focused on the old tab, so switch to the new window handle.
-    const newTab = (await this.driver.getAllWindowHandles())[1];
-    await this.driver.switchTo().window(newTab);
+  // Switch back to web content context.
+  await this.driver.setContext(firefox.Context.CONTENT);
```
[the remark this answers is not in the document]
rhelmer: Yes, as long as we leave it set to whatever we've told the caller they can expect in the jsdoc is fine :)

Dexterp37 (approved):
r+ with the changelog file adjusted (the added line needs to be moved under "unreleased", automerge added it in 0.8)

PR #315, opened 2021-02-01
Remove the mirrored repo from the release process

### LINKED ISSUE #302: Fix tags in our release process
The Addons team pointed out that our tags in mozilla-extensions/rally-core-addon point to the wrong commit id (not the one releases are built from). This means our [release process](https://github.com/mozilla-rally/core-addon/blob/master/docs/RELEASE_PROCESS.md#finish-a-release-branch), specifically step 8 (`git push mozext v25.0.0`) is wrong: that's probably because we're pushing tags across repositories.

The tag should point to the last commit on the `release` branch (as we do in this repository).

A potential solution to this problem would be to change the [release process](https://github.com/mozilla-rally/core-addon/blob/master/docs/RELEASE_PROCESS.md#finish-a-release-branch) so that `git push mozext v25.0.0` in  step 8 becomes instead something like step 4 (same paragraph) in the `mozilla-extensions` repo. This basically means we'd be creating the tags manually in the two repos, but at the same points.

@rhelmer does Shipit allow picking commits from branches other than `master`?

Dexterp37:
Fixes #302 

We're in the process of decommissioning the mirrored repository in mozilla-extensions/rally-core-addon.
This PR simplifies the release process to remove the now unneded steps releted to the mirrored repository.

Checklist for reviewer:

### DIFF
--- a/docs/RELEASE_PROCESS.md
+++ b/docs/RELEASE_PROCESS.md
@@ -70,27 +70,15 @@ When CI has finished and is green for your specific release branch, you are read
     1. [Draft a New Release](https://github.com/mozilla-rally/core-addon/releases/new) in the GitHub UI (`Releases > Draft a New Release`).
     2. Enter `v<myversion>` as the tag. It's important this is the same as the version you specified to the `prepare_release.sh` script, with the `v` prefix added.
     3. Select the `release` branch as the target.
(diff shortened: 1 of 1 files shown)


Dexterp37:
This already happened in #322

PR #337, opened 2021-02-05
Reject in rally.js `initialization`

### LINKED ISSUE #321: Make `rally.initialize()` reject if the core addon is not there
We currently have no way in `rally.js` to tell the studies using it that the study should not run (e.g. if the study addon is installed but no core-addon is present/user is not enrolled).

Since studies are **required** to call `rally.initialize`, we should probably make that function reject to inform studies that they should not continue.

Another approach would be to provide another API to notify the study that it can start.

Dexterp37:
Fixes #321

Note that this will need to be rebased after #334 is merged.

Checklist for reviewer:

### DIFF
--- a/support/CHANGELOG.md
+++ b/support/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased changes
 
+* [#337](https://github.com/mozilla-rally/rally-core-addon/pull/337): `initialize` rejects if the study is not allowed to start.
+
 # v0.2.0 (2021-02-02)
 
 * [#316](https://github.com/mozilla-rally/rally-core-addon/pull/316): Migrate to ES6 modules.
--- a/support/rally.js
+++ b/support/rally.js
@@ -37,19 +37,27 @@ export default class Rally {
     this._key = key;
     this._enableDevMode = Boolean(enableDevMode);
 
-    await this._checkRallyCore().then(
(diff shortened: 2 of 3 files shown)


rollup.config.addon.js
```diff
@@ -15,11 +15,17 @@ export default (cliArgs) => {
...
     plugins: [
       replace({
-        __ION_STUDIES_LIST__: cliArgs['config-study-list-url'] ?
-          `'${cliArgs['config-study-list-url']}'` :
-          "'https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/pioneer-study-addons-v1/records'",
+        // NOTE: if this URL ever changes, you will have to update the domain in
+        // the permissions in manifest.json.
+        __STUDIES_LIST__: cliArgs['config-studies-list-url'] ?
+          `'${cliArgs['config-studies-list-url']}'` :
+          "'https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/rally-studies-v1/records'",
```
rhelmer, line 22: Ah thanks for catching that, unfortunate that it needs to be duplicated... we don't *need* this right now since we support remote-settings but I think keeping it working is good since it's what we use for `build-local-addon` and eventually down the road we'll probably want to go unprivileged so this is good.

rhelmer (approved):

PR #344, opened 2021-02-05
add page + functionality to handle non-US users in the Core Add-On

### LINKED ISSUE #297: Implement a page to show if outside of the U.S.
Rally is currently not supported outside of the U.S.. We can detect this from the core-addon (see #263).

This bug is for designing and implementing the page to point users outside of the U.S. to. Ryan provided the mock:

image

@hamilton this is a blocker for GTM. Is this something you can take?

hamilton:
Closes #297 

This PR does the following:
- implements the page telling non-US users that they can't sign up for Rally. This uses `store.updatePlatformEnrollment(false)`
- ~adds an API to the Core add-on called `uninstall-rally` that we can use to uninstall the Core Add-On, used on the non-US user splash page. I'm open to remove this API call; it feels a little bit risky to have an API like this.~
- ~adds a unit test for the `uninstall-rally` API call.~

Checklist for reviewer:


src/App.svelte
```diff
@@ -40,12 +41,12 @@
...
-            // reset to first run until Rally uninstalls itself.
-            firstRun = true;
           }}
         />
     {/if}
   {/if}
 {:else}
-  <div>Sorry, Rally is not supported in this locale.</div>
+  <NonUSSplashPage on:remove-extension={() => {
+    store.uninstallRally();
```
Dexterp37, line 50:
```suggestion
    store.updatePlatformEnrollment(false);
```

Can you kindly check if this works?

src/App.svelte
```diff
@@ -40,12 +41,12 @@
       <Main
           on:leave-rally={() => {
             store.updatePlatformEnrollment(false);
-            // reset to first run until Rally uninstalls itself.
-            firstRun = true;
```
Dexterp37, line 44: Why was this removed?

core-addon/Core.js
```diff
@@ -194,6 +194,11 @@ module.exports = class Core {
         return this._unenrollStudy(message.data.studyID);
       case "unenrollment":
         return this._unenroll();
+      case "uninstall-rally":
+        // If a user is uninstalling Rally because they do not qualify to be 
+        // on the platform (e.g. they are non-US), uninstall it for them on request.
+        browser.management.uninstallSelf({ showConfirmDialog: false });
+        return Promise.resolve();
```
Dexterp37, line 201:
```suggestion
        return browser.management.uninstallSelf({ showConfirmDialog: false });
```
That API returns a promise, we should return it back to the caller.

Dexterp37 (changes requested):
Stealing review!

core-addon/Core.js
hamilton: So I had this originally, but it caused the unit test I added to timeout. Any ideas there?

Dexterp37: You probably need [this line](https://github.com/mozilla-rally/rally-core-addon/blob/3a145ca14bc141eb5bb82898e7508c04ecf61aeb/tests/core-addon/unit/Core.test.js#L222). I'd recommend going with what I suggested you in the other comment, it will save you from the need of writing the unit test at all :D

hamilton: Yeah, I think the test already has that line but still times out. It's a problem with the `browser.management.uninstallSelf` mock itself. Maybe they've made it so it really simulates what it does & never actually resolves! At any rate, if `updatePlatformEnrollment(false)` can be overloaded to just follow the uninstall path, that works for me too!

src/web-extension-api.js
```diff
@@ -26,7 +26,7 @@ async function sendToCore(port, type, payload) {
     "study-enrollment",
     "study-unenrollment",
     "unenrollment",
-    "update-demographics",
+    "update-demographics"
```
Dexterp37, line 29: Can you kindly retain the trailing comma, here?

src/routes/non-eligible-splashes/NonUSUser.svelte
```diff
@@ -0,0 +1,91 @@
...
+                </div>
+                <div class="waitlist">
+                <a target="_blank" href="__BASE_SITE__/waitlist/">Join the Waitlist
+                    <ExternalLink /></a>
+                </div>
+            </div>
+        </section>
+    </Main>
+</Layout>
+{/if}
```
Dexterp37, line 91: nit: can you add a newline?

Dexterp37 (approved):
r+wc

Issue #352, opened 2021-02-09
Deal with consent and side-loaded studies


Dexterp37:
The proposal lives here: https://docs.google.com/document/d/1E2HUP07WCdlj9I-tH9PoCUU3cOrc0JWjEw2oxnpszmE/edit#

PR #386, opened 2021-02-12
add remote-controllable "pause" state for studies

rhelmer:
Checklist for reviewer:

### DIFF
--- a/core-addon/Core.js
+++ b/core-addon/Core.js
@@ -248,9 +248,11 @@ module.exports = class Core {
    */
   async _handleExternalMessage(message, sender) {
     // We only expect messages coming from known installed studies.
-    let installedStudies = (await this._availableStudies)
+    const availableStudies = await this._availableStudies;
+    let installedStudies = availableStudies
       .filter(s => s.studyInstalled)
       .map(s => s.addonId);
+
     if (!installedStudies.includes(sender.id)) {
       throw new Error(`Core._handleExternalMessage - unexpected sender ${sender.id}`);
     }
@@ -259,6 +261,10 @@ module.exports = class Core {
     if (!joinedStudies.includes(sender.id)) {
       throw new Error(`Core._handleExternalMessage - ${sender.id} not joined`);
     }
+
+    const pausedStudies = availableStudies
+      .filter(s => s.studyPaused)
+      .map(s => s.addonId);
 
     switch (message.type) {
       case "core-check": {
@@ -271,6 +277,9 @@ module.exports = class Core {
         };
       }
       case "telemetry-ping": {
+        if (pausedStudies.includes(sender.id)) {
+          throw new Error(`Core._handleExternalMessage - ${sender.id} is paused and may not send data`);
+        }
         const { payloadType, payload, namespace, keyId, key } = message.data;
         let rallyId = await this._storage.getRallyID();
         return await this._dataCollection.sendPing(
@@ -487,6 +496,8 @@ module.exports = class Core {
    */
   async _sendMessageToStudy(studyId, type, payload) {
     const VALID_TYPES = [
+      "pause",
+      "resume",
       "uninstall",
     ];
 
@@ -515,6 +526,7 @@ module.exports = class Core {
 
   /**
    * Update the `studyInstalled` property for the available studies.
+   * If any studies should be disabled or enabled, then do so now.
    *
    * @returns {Promise(Array<Object>)} resolved with an array of studies
    *          objects, or an empty array on failures. Each study object
@@ -530,10 +542,38 @@ module.exports = class Core {
       await browser.management.getAll().then(addons =>
         addons.filter(a => a.type == "extension")
           .map(a => a.id));
+
+    // Attempt to resume any paused studies, or pause any running
+    // studies, as appropriate.
+    await this._sendRunState(studies, installedAddonsIds);
+
     return studies.map(s => {
       s.studyInstalled = installedAddonsIds.includes(s.addonId);
       return s;
     });
+  }
+
+  /**
+   * Send run state (paused, running) message to study add-on(s).
+   *
+   * @param {Array} studies - list of available studies
+   * @param {Array} installedAddonsIds - list of installed study add-on IDs
+   */
+
+  async _sendRunState(studies, installedAddonsIds) {
+    for (const study of studies) {
+      if (installedAddonsIds.includes(study.addonId)) {
+        try {
+          if (study.studyPaused) {
+            await this._sendMessageToStudy(study.addonId, "pause", {});
+          } else {
+            await this._sendMessageToStudy(study.addonId, "resume", {});
+          }
+        } catch (err) {
+          console.error("Changing study state failed:", err);
+        }
+      }
+    }
   }
 
   /**
@@ -603,15 +643,17 @@ module.exports = class Core {
   async _sendStateUpdateToUI() {
     let enrolled = !!(await this._storage.getRallyID());
     let firstRunCompleted = !!(await this._storage.getFirstRunCompletion());
+
     let availableStudies = await this._availableStudies;
     let demographicsData = await this._storage.getDemographicsData();
 
     // Report a study as joined only if consent was given.
+    // Filter out any paused studies, unless the user is currently enrolled, so they can still leave a study even if paused.
     let joinedStudies = await this._storage.getActivatedStudies();
     availableStudies = availableStudies.map(s => {
       s.studyJoined = joinedStudies.includes(s.addonId);
       return s;
-    });
+    }).filter(study => (!study.studyPaused || study.studyJoined));
 
     const newState = {
       enrolled,
--- a/support/rally.js
+++ b/support/rally.js
@@ -5,6 +5,11 @@
 const CORE_ADDON_ID = "rally-core@mozilla.org";
 const SIGNUP_URL = "https://rally.mozilla.org/rally-required";
 
+const STATES = {
+	RUNNING: "running",
+	PAUSED: "paused",
+}
+
 export default class Rally {
   /**
    * Initialize the Rally library.
@@ -27,11 +32,24 @@ export default class Rally {
    *        Whether or not to initialize Rally.js in developer mode.
    *        In this mode we ignore problems when communicating with
    *        core add-on.
+   *
+   * @param {Function} stateChangeCallback
+   *        A function to call when the study is paused or running.
+   *        Takes a single parameter, `message`, which is the {String}
+   *        received regarding the current study state ("paused" or "running".)
    */
-  async initialize(keyId, key, enableDevMode) {
+  async initialize(keyId, key, enableDevMode, stateChangeCallback) {
     console.debug("Rally.initialize");
 
     this._validateEncryptionKey(keyId, key);
+
+    if (!stateChangeCallback) {
+      throw new Error("Rally.initialize - Initialization failed, stateChangeCallback is required.")
+    }
+
+    if (typeof stateChangeCallback !== "function") {
+      throw new Error("Rally.initialize - Initialization failed, stateChangeCallback is not a function.")
+    }
 
     this._keyId = keyId;
     this._key = key;
@@ -62,6 +80,11 @@ export default class Rally {
     // Listen for incoming messages from the Core addon.
     browser.runtime.onMessageExternal.addListener(
       (m, s) => this._handleExternalMessage(m, s));
+
+
+    // Set the initial state to running, and register callback for future changes.
+    this._state = STATES.RUNNING;
+    this._stateChangeCallback = stateChangeCallback;
 
     // We went through the whole init process, it's now safe
     // to use the Rally public APIs.
@@ -94,6 +117,26 @@ export default class Rally {
   }
 
   /**
+   * Pause the current study.
+   */
+  _pause() {
+    if (this._state !== STATES.PAUSED) {
+      this._stateChangeCallback("pause");
+      this._state = STATES.PAUSED;
+    }
+  }
+
+  /**
+   * Resume the current study, if paused.
+   */
+  _resume() {
+    if (this._state !== STATES.RUNNING) {
+      this._stateChangeCallback("resume");
+      this._state = STATES.RUNNING;
+    }
+  }
+
+  /**
    * Handles messages coming in from external addons.
    *
    * @param {Object} message
@@ -113,6 +156,12 @@ export default class Rally {
     }
 
     switch (message.type) {
+      case "pause":
+        this._pause();
+        break;
+      case "resume":
+        this._resume();
+        break;
       case "uninstall":
         return browser.management.uninstallSelf({showConfirmDialog: false});
       default:
@@ -173,6 +222,12 @@ export default class Rally {
         `Rally.sendPing - Developer mode. ${payloadType} will not be submitted`,
         payload
       );
+      return;
+    }
+
+    // When paused, not send data.
+    if (this._state === STATES.PAUSED) {
+      console.debug("Rally.sendPing - Study is currently paused, not sending data");
       return;
     }
 
--- a/tests/core-addon/unit/Core.test.js
+++ b/tests/core-addon/unit/Core.test.js
@@ -34,6 +34,10 @@ describe('Core', function () {
       .resolves();
     chrome.management.getAll.yields(
       [{type: "extension", id: FAKE_STUDY_ID}]);
+    browser.storage.local.get
+      .callsArgWith(1, {activatedStudies: [FAKE_STUDY_ID]})
+      .resolves();
+    chrome.runtime.sendMessage.yields();
 
     // NodeJS doesn't support "fetch" so we need to mock it
     // manually (or use a third party package). This isn't too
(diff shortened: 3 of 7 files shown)


rhelmer:
@Dexterp37 there are two commits in this PR - the oldest shows the "disable study add-on with privileged API" approach, and the more recent commit reverts these and sets up message-passing between `rally.js` and the core-addon.

The overall concept is:

1. on add-on startup and when new data arrives, check `studyPaused` field for each study
2. keep the `studyPaused` state in the core add-on and also fire a message to any installed studies
3. do not allow data transmission for "paused" studies (similar to dev mode)

I am imagining study add-ons only need to interact with it like:

```js
rally.addStateListener((state) => { console.debug("fires on pause/resume"));
```

It would be up to the study add-on to actually do the right thing when the callback fires :)

Assuming that all sounds reasonable, I need to sort tests out but wanted to open it up to feedback sooner. I think the problem is that fetching studies happens on every data update now, and the mocks are incomplete.

core-addon/Core.js
```diff
@@ -493,6 +497,23 @@ module.exports = class Core {
...
+            await this._sendMessageToStudy(study.addonId, "pause", {});
+          } else {
+            await this._sendMessageToStudy(study.addonId, "resume", {});
+          }
+        }
+      }
+    } catch (err) {
+      console.error("Changing study state failed:", err);
+    }
+
```
Dexterp37, line 516: Can we break this to a separate function?

support/rally.js
```diff
@@ -84,15 +85,31 @@ export default class Rally {
         await browser.runtime.sendMessage(CORE_ADDON_ID, msg, {});
 
       if (!response
-          || response.type !== "core-check-response"
-          || response.data.enrolled !== true) {
+        || response.type !== "core-check-response"
+        || response.data.enrolled !== true) {
```
Dexterp37, line 89: Mh, ok, what's up with these changes here? Why is your IDE unhappy with our rules? Do we need to change our rules? :)

support/rally.js
```diff
@@ -113,8 +130,14 @@ export default class Rally {
...
     switch (message.type) {
+      case "pause":
+        this._pause();
+        break;
+      case "resume":
+        this._resume();
+        break;
       case "uninstall":
-        return browser.management.uninstallSelf({showConfirmDialog: false});
+        return browser.management.uninstallSelf({ showConfirmDialog: false });
```
Dexterp37, line 140: Same here :)Why is your IDE unhappy with our rules? Do we need to change our rules? :)

support/rally.js
```diff
@@ -152,6 +175,15 @@ export default class Rally {
...
 
+  /**
+   * A listener which fires when the study is paused or resumed.
+   *
+   * @param {function} callback
+   */
+  addStateListener(callback) {
+    this._listener = callback;
+  }
+
```
Dexterp37, line 186: We should make this a mandatory parameter of `initialize`, otherwise folks will forget to call it :)

tests/core-addon/unit/Core.test.js
```diff
@@ -30,10 +30,10 @@ describe('Core', function () {
     // Note that this will fake our data to make FAKE_STUDY_ID look
     // installed.
     chrome.management.getAll
-      .callsArgWith(0, [{type: "extension", id: FAKE_STUDY_ID}])
+      .callsArgWith(0, [{ type: "extension", id: FAKE_STUDY_ID }])
       .resolves();
     chrome.management.getAll.yields(
-      [{type: "extension", id: FAKE_STUDY_ID}]);
+      [{ type: "extension", id: FAKE_STUDY_ID }]);
```
Dexterp37, line 36: Most of the changes to this file are unrelated. Even if our linter is wrong, would you mind reverting it and just keep the relevant changes here?

tests/support/rally.test.mjs
```diff
@@ -63,18 +63,42 @@ describe('Rally', function () {
...
+  describe('_resume()', function () {
+    it('resumes when receiving message', async function () {
+      this.rally._handleExternalMessage("resume", "rally-core@mozilla.org");
+
+      assert.equal(this.rally._state = "running");
+    });
+  });
+
+  /*
+  describe('addStateListener()', function () {
```
Dexterp37, line 95: Don't forget to remove :)

support/rally.js
```diff
@@ -176,6 +208,12 @@ export default class Rally {
       return;
     }
 
+    // When paused, log an error and do not send data.
+    if (this._state === "paused") {
+      console.error("Rally.sendPing - Study is currently paused, not sending data");
```
Dexterp37, line 213: Do we really need this to be an `error`? This is not really actionable by users nor developers, or am I missing something?

Dexterp37 (changes requested):
Thanks for doing this. I like the approach, I have two main observations:

- can we please keep the linting changes out of this PR? It makes review a bit more complicated that it should be. We can tackle them separately (and finally figure out if our linting is weird or your IDE is misbehaving!)
- instead of providing a function to add an observer, can we please require it at initialization?

support/rally.js
rhelmer: I suspect our eslint isn't strict enough, it seems fine with or without these changes. Not sure why VSCode isn't using the same exact config if that's the case though?

Anyway happy to just remove these :) I filed an issue on the linting stuff.

support/rally.js
rhelmer: Hm, I would want people who are testing (or normal users) to be able to report it to us. I think `warn` at least is appropriate, but I'm OK with `debug` here .

rhelmer: IIRC Firefox shows `debug` by default in the console but Chrome does not.

On a different note, I am starting to think maybe this check should be in the core add-on code and we shouldn't just expect `rally.js` to do the right thing, that feels safer in the face of a misbehaving add-on.

core-addon/Core.js
rhelmer: I was thinking I could probably shrink this down if I use a more functional style too :) I'll see what I can come up with.

core-addon/Core.js
```diff
@@ -493,6 +496,23 @@ module.exports = class Core {
       await browser.management.getAll().then(addons =>
         addons.filter(a => a.type == "extension")
           .map(a => a.id));
+
+    // Attempt to resume any paused studies, or pause any running
+    // studies, as appropriate.
+    try {
```
Dexterp37, line 502: Should we wrap the  `this._sendMessageToStudy` in the try instead of the whole iteration loop? For example, this would allow us to keep on pausing/resumng other studies even if a message is thrown for an individual study.

src/web-extension-api.js
```diff
@@ -215,6 +215,7 @@ export default {
     switch (message.type) {
       case "update-state": {
         // update the UI.
+        message.data.availableStudies = message.data.availableStudies.filter(study => !study.studyPaused);
```
Dexterp37, line 218: Since we're not changing the UI (other than hiding studies) based on the paused/resumed stated, let's not leak this logic in the UI. Let's do it in [the function to send the updates](https://github.com/mozilla-rally/rally-core-addon/blob/5faf1f355bf25c92189725725d018b0d7c5a69cb/core-addon/Core.js#L567)

support/rally.js
Dexterp37: Oh yeah, indeed. See what I'm doing in #352 , because I'm doing the same :)

Dexterp37 (changes requested):

support/rally.js
rhelmer: > We should make this a mandatory parameter of `initialize`, otherwise folks will forget to call it :)

Hm, do you mean we'd pass the callback function as part of initialize instead of the rally study needing to call `addStateListener()`?

Dexterp37: Yes!

tests/support/rally.test.mjs
rhelmer: Not needed now that we're doing it as part of initialize, I guess :)

rhelmer:
@Dexterp37 thanks for all the feedback so far! The last(?) big problem I am running into is that this breaks the `Core.js` "dispatches enrollment messages" test (and likely others too), `this._sendMessageToStudy()` is hanging and I haven't figured out the right way to mock it.

I have the rallyjs tests working, but I am holding off on integration tests until you land #392. 

Everything looks OK to me in manual testing and from the rally.js perspective unit tests, so I suspect it's just something I'm not doing quite right with the mocks in the Core.js unit tests which I know you are more familiar with :)

support/rally.js
```diff
@@ -176,6 +211,12 @@ export default class Rally {
       return;
     }
 
+    // When paused, not send data.
```
rhelmer, line 214: Note to self - move this from `rally.js` into `Core.js`, so we do not need to trust the (potentially misbehaving) study extension.

tests/support/rally.test.mjs
```diff
@@ -67,14 +68,50 @@ describe('Rally', function () {
...
+  describe('_pause()', function () {
+    it('pauses when receiving message', async function() {
+      chrome.runtime.sendMessage.flush();
+      chrome.runtime.sendMessage.yields();
+
+      await this.rally.initialize(
+        "key-id",
+        {},
+        true, // Developer mode.
+        () => {}
```
Dexterp37, line 88: Can we also test that this callback gets called and with the proper argument?

tests/support/rally.test.mjs
```diff
@@ -67,14 +68,50 @@ describe('Rally', function () {
...
+  describe('_resume()', function () {
+    it('resumes when receiving message', async function () {
+      chrome.runtime.sendMessage.flush();
+      chrome.runtime.sendMessage.yields();
+
+      await this.rally.initialize(
+        "key-id",
+        {},
+        true, // Developer mode.
+        () => {},
```
Dexterp37, line 106: Can we also test that this callback gets called and with the proper argument?

tests/support/rally.test.mjs
```diff
@@ -28,6 +28,7 @@ describe('Rally', function () {
         "key-id",
         {},
         true, // Developer mode.
+        () => {},
```
Dexterp37, line 31: Can we also add a new test here to make sure that we throw if no callback is provided?

core-addon/Core.js
```diff
@@ -493,12 +496,40 @@ module.exports = class Core {
       await browser.management.getAll().then(addons =>
         addons.filter(a => a.type == "extension")
           .map(a => a.id));
+
+    // Attempt to resume any paused studies, or pause any running
+    // studies, as appropriate.
+    await this._sendRunState(studies, installedAddonsIds);
```
Dexterp37, line 502: Would you kindly add unit tests for `_sendRunState`?

core-addon/Core.js
```diff
@@ -493,12 +496,40 @@ module.exports = class Core {
...
   }
 
+  /**
+   * Send run state (paused, running) message to study add-on(s).
+   *
+   * @param {Array} studies - list of available studies
+   * @param {Array} installedAddonsIds - list of installed study add-on IDs
+   */
+
+  async _sendRunState(studies, installedAddonsIds) {
```
Dexterp37, line 517: Are you planning on adding an integration test for this? I'm ok with merging this without an integration test as long as we do in depth manual testng :)

Dexterp37 (changes requested):
This looks good overall, I left a few nits. I'd love to have one final look at this after it is rebased on the latest master.

Dexterp37:
@rhelmer note that the test is failing because `_sendStateUpdateToUI` is awaiting on [this._availableStudies](https://github.com/mozilla-rally/rally-core-addon/blob/9fdbcb120e74d1a25487bf0b22672aa20e93c384/core-addon/Core.js#L597) (which is ok). To make the test pass, we can simply mock that function like this `this.core._sendStateUpdateToUI = async () => {};` , right after [this line](https://github.com/mozilla-rally/rally-core-addon/blob/9fdbcb120e74d1a25487bf0b22672aa20e93c384/tests/core-addon/unit/Core.test.js#L172)

core-addon/Core.js
rhelmer: Yes, honestly I expect the integration test to be fairly simple, but I want to rebase on top of the improvements you made there recently before I start.

tests/support/rally.test.mjs
rhelmer: I meant to bring this up, thanks... since we're using positional arguments (as opposed to an options object or parameter desctructuring etc) then it seems like we'd need to enforce all of the options (throw if any argument is not provided or an incorrect type)

rhelmer:
> @rhelmer note that the test is failing because `_sendStateUpdateToUI` is awaiting on [this._availableStudies](https://github.com/mozilla-rally/rally-core-addon/blob/9fdbcb120e74d1a25487bf0b22672aa20e93c384/core-addon/Core.js#L597) (which is ok). To make the test pass, we can simply mock that function like this `this.core._sendStateUpdateToUI = async () => {};` , right after [this line](https://github.com/mozilla-rally/rally-core-addon/blob/9fdbcb120e74d1a25487bf0b22672aa20e93c384/tests/core-addon/unit/Core.test.js#L172)

This worked great, thanks! I went ahead and rebased both `master` and my branch on my github fork, overall works OK but now there are two new test failures for telemetry tests that were modified in the meantime :) I'll work on that and also adding integration tests today.

support/rally.js
```diff
@@ -27,8 +32,11 @@ export default class Rally {
    *        Whether or not to initialize Rally.js in developer mode.
    *        In this mode we ignore problems when communicating with
    *        core add-on.
+   *
+   * @param {Function} stateChangeCallback
+   *        A function to call when the study is paused or running.
```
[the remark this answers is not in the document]
rhelmer: I meant to bring this up, thanks... since we're using positional arguments (as opposed to an options object or parameter desctructuring etc) then it seems like we'd need to enforce all of the options (throw if any argument is not provided or an incorrect type)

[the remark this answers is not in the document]
rhelmer: Well, my previous comment assumes we want to insist that these are required and not optional arguments. If they are optional, then making sure they are the expected type and such is fine and I'll just do that for now.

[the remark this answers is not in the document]
rhelmer: Oh, it turns out that jsdoc has support for callback functions: https://jsdoc.app/tags-callback.html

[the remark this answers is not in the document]
rhelmer: Ugh, it turns out that the jsdoc `@callback` needs to be registered elsewhere in the file. It's probably just fine for auto-generated docs, but since it won't be above the initialize function here I don't think it's useful. I went ahead and documented it in the jsdoc header for initialize, instead.

core-addon/Core.js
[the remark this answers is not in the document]
rhelmer: I think this is a good idea but I ran out of time to do that, should we defer to a separate issue? I'd like to think about this a bit more holistically, I feel like we are mixing styles and it'd be helpful to either go more or less consistently functional :)

Dexterp37 (changes requested):
Last batch of small tweaks. I'll happily r+ once the tests pass!

support/rally.js
```diff
@@ -93,6 +116,22 @@ export default class Rally {
     }
   }
 
+  /**
+   * Pause the current study.
+   */
+  _pause() {
+    this._stateChangeCallback("pause");
+    this._state = STATES.PAUSED;
```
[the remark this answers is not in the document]
rhelmer: My feeling is that checking to see if the study is paused is additional message overhead (the core only knows that the study *should* be paused, we need to actually send a message to the study to determine its state.)

Sending a single message to the study informing it that it should pause makes sense in this context, since only the study has this state. I might be wrong though I'll think about this a bit more :)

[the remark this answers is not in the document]
Dexterp37: I mean just

```suggestion
  _pause() {
    if (this._state !== STATES.PAUSED) {
    this._stateChangeCallback("pause");
    this._state = STATES.PAUSED;
    }
```

just to make sure studies don't pause twice if we make mistakes in the core addon and send the message twice. The rationale is that we control the quality of the code in the core add-on, but we can't guarantee that a double pause won't create problems in the study right now (

support/rally.js
rhelmer: Hm. Actually I'd like to defer this one, I think we don't have the information we need. I'll send you a message about this as it is a little involved.

support/rally.js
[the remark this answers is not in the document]
rhelmer: Oh shoot right for some reason I thought you were pointing at the `Core.js` and not `rally.js` :) of course this sounds good ty.

core-addon/Core.js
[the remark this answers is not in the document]
rhelmer: https://github.com/mozilla-rally/rally-core-addon/issues/420

core-addon/Core.js
```diff
@@ -271,6 +277,9 @@ module.exports = class Core {
         };
       }
       case "telemetry-ping": {
+        if (pausedStudies.includes(sender.id)) {
+          throw new Error(`Core._handleExternalMessage - ${sender.id} is paused and may not send telemetry`);
```
Dexterp37, line 281:
```suggestion
          throw new Error(`Core._handleExternalMessage - ${sender.id} is paused and may not send data`);
```

Let's try not to use the term telemetry, as this might get confused with Firefox Telemetry.

Dexterp37 (approved):
r+ with a small nit

PR #422, opened 2021-02-19
removes terser from svelte rollup configuration

### LINKED ISSUE #330: Make the generated code more stable
This was reported by the Add-ons team. Fixing this would make reviews a bit simpler. For example, comparing `public/build/bundle.js` between version 0.7.1 and 0.8.0 includes many variable name changes and one letter variables:

image

hamilton:
closes #330 

Checklist for reviewer:


rollup.config.js
```diff
@@ -94,10 +93,6 @@ export default (cliArgs) => [{
     // Watch the `public` directory and refresh the
     // browser on changes when not in production
     !production && livereload("public"),
-
-    // If we're building for production (npm run build
-    // instead of npm run dev), minify
-    production && terser(),
```
Dexterp37, line 100: What if, instead of removing the plugin, we only disabled mangling?

```js
    // If we're building for production (npm run build
    // instead of npm run dev), minify
    production && terser({
      mangle: false,
    }),
```

Dexterp37: See my comment [here](https://github.com/mozilla-rally/rally-core-addon/issues/330#issuecomment-781966238)

hamilton: Yeah, I didn't perform any due dilligence here :) much better strategy. I'll add that option.

Issue #405, opened 2021-02-17
ensure Rally deletion pings and other cleanup happen on uninstall


rhelmer:
If Rally is uninstalled from within the options page UI, then the `Core._unenroll()` method is called: https://github.com/mozilla-rally/rally-core-addon/blob/d72dd24da2bd41451e4236f59231d32d60625bb7/core-addon/Core.js#L429

However, users can uninstall via `about:addons` or even just context-click on the toolbar icon -> Remove Extension, which will cause the core extension to be removed without the chance to run this sort of unenrollment code.

The one API available to do anything post-uninstall is [browser.runtime.setUninstallURL](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/setUninstallURL), but that only allows registering a URL to open in a new tab, and that's not sufficient for the things we need to do here such as sending deletion pings.

Since the core extension is privileged, we could additionally register an `onUninstall` handler in a privileged API, similar to what we do here with `onShutdown`: https://github.com/mozilla-rally/rally-core-addon/blob/d72dd24da2bd41451e4236f59231d32d60625bb7/core-addon/FirefoxPrivilegedApi.js#L75-L77

rhelmer:
A corner-case that I think we should explicitly avoid trying to handle is where the XPI is deleted from the filesystem while Firefox is shutdown. This isn't a supported way to remove extensions, although we do see it occasionally from malware and anti-virus software. The problem with trying to handle this case is that the code (including any shutdown handler etc) is no longer present on the filesystem, and Firefox will then remove any metadata about the extension on startup.

Dexterp37:
You're right Rob, that's a weird edge case. In addition to the privileged fix, we should think on how this would work cross-browser. The only API we could leverage, as you mentioned, is `setUninstallURL`. That reports this in [the docs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/setUninstallURL):

`Sets the URL to be visited when the extension is uninstalled. This may be used to clean up server-side data, do analytics, or implement surveys. The URL can be a maximum 255 characters.`

One potential way to address this would be to add Glean.js to the website too and have the URL opened by `setUninstallURL` to send a deletion-request ping that contains the rally id for the user who is leaving. However, while this seems to be technically feasible, my concern would be sending the rally id over the wire.

[...]

PR #453, opened 2021-02-23
set deletion ID in uninstallURL, for issue #405

rhelmer:
Checklist for reviewer:

### DIFF
--- a/core-addon/Core.js
+++ b/core-addon/Core.js
@@ -7,6 +7,10 @@ const DataCollection = require("./DataCollection.js");
 
 // The path of the embedded resource used to control options.
 const OPTIONS_PAGE_PATH = "public/index.html";
+// The website to post deletion IDs to.
+const OFFBOARD_URL = "https://production.rally.mozilla.org/offboard";
+// The static website to send offboarded users to (those with no deletion ID).
+const LEAVE_URL = "__BASE_SITE__/leaving-rally";
 
 module.exports = class Core {
   /**
@@ -42,9 +46,27 @@ module.exports = class Core {
     this._connectionPort = null;
   }
 
+  /**
+   * Set the URL to be opened in a new tab when the core add-on is uninstalled.
+   *
+   * After enrollment, a deletion ID will be available to provide to this URL.
+   * However, if the user uninstalls the core add-on without enrolling first, then
+   * this will open the offboarding URL without the deletion ID.
+   */
+  async setUninstallURL() {
+    // set the URL to redirect when a user uninstalls Rally.
+    const deletionId = await this._storage.getDeletionID();
+    if (deletionId) {
+      // if enrolled,include the deletion ID, for deleting data without exposing the Rally ID.
+      browser.runtime.setUninstallURL(`${OFFBOARD_URL}?id=${deletionId}`);
+    } else {
+      browser.runtime.setUninstallURL(LEAVE_URL);
+    }
+  }
+
   initialize() {
-    // set the URL to redirect when a user uninstalls Rally
-    browser.runtime.setUninstallURL("__BASE_SITE__/leaving-rally");
+    this.setUninstallURL();
+
     // Whenever the addon icon is clicked, open the control page.
     browser.browserAction.onClicked.addListener(this._openControlPanel);
     // After installing the addon, make sure to show the control page.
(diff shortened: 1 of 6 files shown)


rhelmer:
I don't know if we want to merge this in a hurry, but it addresses https://github.com/mozilla-rally/rally-core-addon/issues/405#issuecomment-784415000 which I think is probably our only avenue for approaching this, unfortunately.

Dexterp37 (changes requested):
The PR looks good (it requires a changelog entry).

I'm requesting changes just to prevent it from being merged, as this needs discussion about the implications.

rhelmer:
> The PR looks good (it requires a changelog entry).
> 
> I'm requesting changes just to prevent it from being merged, as this needs discussion about the implications.

Added changelog entry, but not re-requesting review per above.

rhelmer:
@Dexterp37 I've added a separate commit, to match the in-progress proposal (linked in the commit message). Normally I'd let the proposal run its course first but given that this is an alpha-blocker I'd like to get a jump on it.

rhelmer:
@Dexterp37 one thing I don't like here is that the name `_sendEmptyPing`is now a lie :) since we send a deletion ID only for platform enrollment.

I don't *think* there's any other case where we need to send the deletion ID, but I may be missing something.

core-addon/Core.js
```diff
@@ -42,9 +42,20 @@ module.exports = class Core {
     this._connectionPort = null;
   }
 
+  async setUninstallURL() {
+    // set the URL to redirect when a user uninstalls Rally.
+    const deletionId = await this._storage.getDeletionID();
```
Dexterp37, line 47: Why do we expect the ID to be unavailable? Can you document our expectations here?

core-addon/DataCollection.js
```diff
@@ -26,16 +26,20 @@ module.exports = class DataCollection {
    * @param {String} namespace
    *        The namespace to route the ping. This will define the
    *        `schemaNamespace` and `studyName` properties of the ping.
+   * @param {String} deletionId
+   *        It's sent in the ping, if present, to track deletion of user data without exposing the Rally ID.
    */
-  async _sendEmptyPing(rallyId, payloadType, namespace) {
+  async _sendEmptyPing(rallyId, payloadType, namespace, deletionId) {
```
Dexterp37, line 32: We should probably rename this. What about `_sendPingWithDeletionId`?

Dexterp37 (commented):
We should also document the data documentation under docs/ and request a data review.

core-addon/Core.js
rhelmer: I'll add a comment, but it's first set during enrollment so if you have the core add-on installed but have not enrolled then there is neither enrollment nor deletion pings set, yet. In this case we still want to show the offboarding URL on add-on uninstall, but there's no data to delete (please correct if I am misunderstanding though)

rhelmer:
> We should also document the data documentation under docs/ and request a data review.

Hm, good point... reading https://github.com/mozilla-rally/rally-core-addon/blob/master/docs/DATA_COLLECTION.md I suspect that we might want to make a few other small changes like `encryptionKeyId` maybe shouldn't be "discarded" if we're going to be sending data there after all?

I'd also like to give data pipeline folks time to comment on and accept the proposal since it's possible we need other modifications, I'm feeling pretty confident that this will not have an impact on the rest of this patch though.

core-addon/Core.js
Dexterp37: Makes sense :) Let's just leave a comment so that we don't forget in N months ;)

Dexterp37:
> 
> 
> > We should also document the data documentation under docs/ and request a data review.
> 
> Hm, good point... reading https://github.com/mozilla-rally/rally-core-addon/blob/master/docs/DATA_COLLECTION.md I suspect that we might want to make a few other small changes like `encryptionKeyId` maybe shouldn't be "discarded" if we're going to be sending data there after all?

Yes, we'll need to use the proper key (it's already in the code, we need to double check if it's being used for enrollment/deletion). At the very least, we should fix the docs :)

> I'd also like to give data pipeline folks time to comment on and accept the proposal since it's possible we need other modifications, I'm feeling pretty confident that this will not have an impact on the rest of this patch though.

Yes, makes sense. I finalized the proposal @knowtheory started today, I suspect we should be unblocked within the next 48 hours at most.

core-addon/Core.js
```diff
@@ -7,6 +7,10 @@ const DataCollection = require("./DataCollection.js");
 
 // The path of the embedded resource used to control options.
 const OPTIONS_PAGE_PATH = "public/index.html";
+// The website to post deletion IDs to.
+const OFFBOARD_URL = "https://production.rally.mozilla.org/offboard";
+// The static website to send offboarded users to (those with no deletion ID).
+const LEAVE_URL = "https://rally.mozilla.org/leaving-rally/";
```
Dexterp37, line 13:
```suggestion
const LEAVE_URL = "__BASE_SITE__/leaving-rally";
```
I think this can probably remain based on the provided base site.

core-addon/Core.js
```diff
@@ -367,14 +389,19 @@ module.exports = class Core {
...
+    const rallyId = await browser.firefoxPrivilegedApi.generateUUID();
+    const deletionId = await browser.firefoxPrivilegedApi.generateUUID();
+
+    // Store IDs locally for future use.
+    await this._storage.setRallyID(rallyId);
+    await this._storage.setDeletionID(rallyId);
 
-    // Store it locally for future use.
-    await this._storage.setRallyID(uuid);
+    // override the uninstall URL to include the rallyID, for deleting data without exposing the Rally ID.
```
Dexterp37, line 400:
```suggestion
    // Override the uninstall URL to include the rallyID, for deleting data without exposing the Rally ID.
```

nit!

Dexterp37 (approved):
This looks good to me ( I left a few comments about nits). Please kindly take this for a manual test drive to make sure the expexted pings are still generated before merging :-)

rhelmer:
> This looks good to me ( I left a few comments about nits). Please kindly take this for a manual test drive to make sure the expexted pings are still generated before merging :-)

Verified in dev mode, I see enrollment and (for studies) unenrollment pings. Removing the extension directly or via directly uninstalling opens the correct offboarding URL with the correct ID.

Issue #444, opened 2021-02-22
Export the states from rally.js


Dexterp37:
We should export the [`STATES`](https://github.com/mozilla-rally/rally-core-addon/blob/587c233e3d228e9357bdd864dd10b847fcfe427b/support/rally.js#L8-L11) from rally.js so that consumers don't need to know the literal values (e.g. "running") and could just import the named constants.

PR #468, opened 2021-03-01
export rally run states, fixes issue #444

rhelmer:
Checklist for reviewer:

### DIFF
--- a/support/CHANGELOG.md
+++ b/support/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased changes
 
+# v0.5.0 (2021-03-03)
+
 * [#468](https://github.com/mozilla-rally/rally-core-addon/pull/468): export rally run states instead of using strings.
 
 # v0.4.0 (2021-02-19)
--- a/support/CHANGELOG.md
+++ b/support/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased changes
 
+* [#468](https://github.com/mozilla-rally/rally-core-addon/pull/468): export rally run states instead of using strings.
+
 # v0.4.0 (2021-02-19)
 
 * [#402](https://github.com/mozilla-rally/rally-core-addon/pull/402): point to the correct URL when no core add-on is found.
--- a/support/package-lock.json
+++ b/support/package-lock.json
@@ -1,5 +1,5 @@
 {
   "name": "@mozilla/rally",
-  "version": "0.3.0",
+  "version": "0.5.0",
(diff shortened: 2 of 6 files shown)


rhelmer:
Maybe a test is overkill :) But I figured it was worth checking that we what we are exporting matches what we expect, this would catch for instance a typo in a string.

rhelmer:
It also doesn't seem very idiomatic to export an upper-case `const` which is why I changed that - I would expect this convention for `const`s that are private but for something exported it should probably have a camel-case name, I think.

Dexterp37 (changes requested):
The code looks good! Would you kindly add a changelog entry to the support library changelog, [here](https://github.com/mozilla-rally/rally-core-addon/blob/master/support/CHANGELOG.md)? I think we want to cut a new release of this too.

Did you test this in the study template? (there's instructions to do so in the README file in the support/ directory)
Do you want to tackle the version bump in this or another PR?

rhelmer:
> The code looks good! Would you kindly add a changelog entry to the support library changelog, [here](https://github.com/mozilla-rally/rally-core-addon/blob/master/support/CHANGELOG.md)? I think we want to cut a new release of this too.

Will do.

> Did you test this in the study template? (there's instructions to do so in the README file in the support/ directory)

I have a patch for that repo that passes tests.

> Do you want to tackle the version bump in this or another PR?

Let's do it here.

rhelmer:
> > Did you test this in the study template? (there's instructions to do so in the README file in the support/ directory)
> 
> I have a patch for that repo that passes tests.

https://github.com/mozilla-rally/study-template/pull/65

support/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased changes
 
+* [#468](https://github.com/mozilla-rally/rally-core-addon/pull/468): export rally run states instead of using strings.
```
Dexterp37, line 3:
```suggestion
# v0.5.0 (2021-03-03)

* [#468](https://github.com/mozilla-rally/rally-core-addon/pull/468): export rally run states instead of using strings.
```

Dexterp37 (approved):
r+ with the nit in the changelog fixed (we should probably have a script for this, too)

rhelmer:
> r+ with the nit in the changelog fixed (we should probably have a script for this, too)

Yes, I think I did not do this because I've only done the core-addon release so far and not totally familiar with the rally.js one :) Good thing we are doing this!

PR #486, opened 2021-03-10
Send the deletion id in the enrollment ping

Dexterp37:
This make sure we're sending both the deletion id and the rally id in the enrollment ping, not two copies of the rally id.

Checklist for reviewer:


rhelmer (approved):
Code changes look good (sorry for missing this!), Arkadiusz confirmed that it looks OK on the server side now. As discussed I'll write a better test for this given what we've learned here.

PR #506, opened 2021-03-18
remove livereload and make sirv quiet, fixes issue 498

rhelmer:
Checklist for reviewer:

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
     "rollup": "^2.42.0",
     "rollup-plugin-copy": "^3.4.0",
     "rollup-plugin-css-only": "^3.1.0",
-    "rollup-plugin-livereload": "^2.0.0",
     "rollup-plugin-postcss": "^3.1.8",
     "rollup-plugin-svelte": "^7.1.0",
     "rollup-plugin-terser": "^7.0.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     "lint-css": "stylelint 'public/*.css' '.storybook/*.css' 'src/**/*.svelte' 'stories/**/*.svelte'",
     "lint-js": "eslint . .storybook",
     "skip-taskcluster": "node -e \"process.exit(!('TASK_ID' in process.env))\"",
-    "start": "sirv public",
+    "start": "sirv public --quiet",
```
rhelmer, line 14: FWIW I don't think this is a super useful target anymore, since the Svelte app only works in webextension contexts. @Dexterp37 I think it might be more useful to have `npm run start` run the webextension w/o watch, wdyt?

Dexterp37: I think we can remove this for now and then add back "start" if we think it's useful. But no strong opinion!

rollup.config.js
```diff
@@ -5,7 +5,6 @@
 import svelte from "rollup-plugin-svelte";
 import resolve from "@rollup/plugin-node-resolve";
 import commonjs from "@rollup/plugin-commonjs";
-import livereload from "rollup-plugin-livereload";
```
Dexterp37, line 8: Would you kindly check if this is being used elsewhere and, if not, would you kindly remove it from the dependencies using [this command](https://docs.npmjs.com/uninstalling-packages-and-dependencies#removing-a-local-package-from-the-packagejson-dependencies)?

Dexterp37 (changes requested):

rollup.config.js
rhelmer: I checked storybook and our normal tests, all seems well. Coming right up!

Dexterp37 (approved):
Tests fail because this PR was not rebased :) Rebase and integration tests will work!

PR #509, opened 2021-03-19
Fix the wrong import in setupTypeScript.js

Dexterp37:
Checklist for reviewer:

### DIFF
--- a/scripts/setupTypeScript.js
+++ b/scripts/setupTypeScript.js
@@ -17,7 +17,7 @@
 */
 
 import fs from "fs";
-import { dirname }, path from 'path';
+import path, { dirname } from "path";
 import { argv } from "process";
 import { fileURLToPath } from 'url';
 
--- a/scripts/setupTypeScript.js
+++ b/scripts/setupTypeScript.js
@@ -17,7 +17,7 @@
 */
 
 import fs from "fs";
-import { dirname, path } from 'path';
+import { dirname }, path from 'path';
 import { argv } from "process";
 import { fileURLToPath } from 'url';
 


scripts/setupTypeScript.js
```diff
@@ -17,7 +17,7 @@
 */
 
 import fs from "fs";
-import { dirname, path } from 'path';
+import { dirname }, path from 'path';
```
brizental, line 20: Hey, are you sure this works? I just tried it in Glean.js and I have to do:

```js
import path from "path";
import { dirname } from "path";
```

Even in Javascript files.

brizental (changes requested):

scripts/setupTypeScript.js
Dexterp37: Well, [yes](https://github.com/mozilla-rally/rally-core-addon/blob/a3052afad8dbd443b40c7edd9ea0da4a64179bdc/tests/core-addon/integration/utils.js#L10)?

brizental: Craziness. Why do I get an error...

brizental (approved):

scripts/setupTypeScript.js
brizental: Ok, I have no documentation link, but I checked it out and it seems:
`import path, { dirname } from "path";` works,
`import { dirname }, path from "path";` doesn't.

Dexterp37, line 20:
```suggestion
import path, { dirname } from "path";
```

Dexterp37:
Merging this, the CI changes are orthogonal and I'll be fixing them in anotehr PR.

PR #532, opened 2021-04-07
log the payload when data collection is disabled, for QA purposes

rhelmer:
Checklist for reviewer:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 [Full changelog](https://github.com/mozilla-rally/core-addon/compare/v1.2.0...master)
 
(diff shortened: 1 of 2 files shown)


rhelmer:
@Dexterp37 I am finding it very difficult to QA studies that haven't been released yet, what do you think about doing this?

Happy to consider alternatives but I'd like to make it as easy as we can for folks doing this sort of testing. If we're concerned about potential leakage I could see dumping the encrypted payload instead but this would be pretty difficult for all but the most sophisticated test user :)

Dexterp37 (approved):
I think this is very reasonable! Please add a changelog entry :)

PR #550, opened 2021-04-19
Do not use extension ID as namespace, closes issue #548

### LINKED ISSUE #548: the core add-on sends the add-on id in the `schemaNamespace` field when it should send the actual namespace
We are not sending the correct `schemaNamespace`, it turns out. The Core add-on is sending this `namespace` field which corresponds to the `studyName` (e.g. `rally-study-01@mozilla.org` for the first Rally study) instead of the actual namespace (e.g. `rally-study-zero-one`).

This will require adding an additional namespace argument in `rally.js` to support this field, probably in the `initialize` method, and then have the `sendPing` method in `rally.js` send that to the Core Add-On [here](https://github.com/mozilla-rally/rally-core-addon/blob/master/support/rally.js#L241-L255) instead of the add-on ID.

One option is to make the API look like the below (e.g. for [`rally-study-01`](https://github.com/mozilla-rally/rally-study-01/blob/master/src/study.js#L24-L32)) – either add a new argument or get rid of the duplicate `keyId` argument and use the `kid` field of the `key`:

```javascript
await rally.initialize(
        "zero-one", // currently the same as the kid field below? is this needed?
        "rally-study-zero-one", // the namespace field we need
        {
          "crv": "P-256",
          "kid": "zero-one",
          "kty": "EC",
          "x": "edhPpqhgK9dD7NaqhQ7Ckw9sU6b39X7XB8HnA366Rjs",
          "y": "GzsfM19n-iH-DVR0iKEoA8BE2CFF46wR__siJ3SdiNs"
        },          
      devMode,
     () => { });
```

cc @Dexterp37 ~ do you have a preference for how we fix this?

rhelmer:
closes #548

Checklist for reviewer:

### DIFF
--- a/support/rally.js
+++ b/support/rally.js
@@ -14,8 +14,8 @@ export class Rally {
   /**
    * Initialize the Rally library.
    *
-   * @param {String} keyId
-   *        The id of the key used to encrypt outgoing data.
+   * @param {String} schemaNamespace
+   *        The namespace for this study. Must match the server-side schema.
    * @param {Object} key
    *        The JSON Web Key (JWK) used to encrypt the outgoing data.
    *        See the RFC 7517 https://tools.ietf.org/html/rfc7517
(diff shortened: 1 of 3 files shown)


support/rally.js
```diff
@@ -191,13 +190,13 @@ export class Rally {
    * @throws {Error} if either the key id or the JWK key object are
    *         invalid.
    */
-  _validateEncryptionKey(keyId, key) {
-    if (typeof keyId !== "string") {
-      throw new Error(`Rally._validateEncryptionKey - Invalid encryption key id ${keyId}`);
+  _validateEncryptionKey(key) {
+    if (typeof key !== "object") {
+      throw new Error("Rally._validateEncryptionKey - Invalid encryption key", key);
```
rhelmer, line 195: I changed this because otherwise this gets converted to `[object Object]`, not super helpful :)

hamilton (approved):
LGTM – thanks!

Dexterp37 (approved):

PR #555, opened 2021-04-21
implements the income question change in the demographic survey

### LINKED ISSUE #496: Implement demographic survey form fields for Income Brackets
See https://github.com/mozilla-rally/rally-core-addon/issues/494 / [this decision brief](https://docs.google.com/document/d/1kUZpaGpaBHF2vA-YnwqAf30WdOy3VE73jRpY5JB_AfE/edit?usp=sharing)

Updated Design replaces the radio buttons to a text field for income entry. 

Figma: https://www.figma.com/file/92ikhj1sHgm3c9jwKp0NHV/Mozilla-Rally-%E2%80%93-Add-on?node-id=2103%3A0 

Expectation is that we'll need to store the specific value entered, but that it may need to be bucketed/categorized to one of the 41 ranges as found in this [brief](https://docs.google.com/document/d/1b2zESmrqhY2uZWMMBArPr-Ux7OKNCj6rqKFz637WJKc/edit#heading=h.6530bmrssqyr)

Additional requirements: 
Screen Shot 2021-03-23 at 10 09 54 AM
- Append a "$" and comma(s) (if required) once focus is removed from the field
- Right align the field 
- Require only numeric entry 

Error States: **_potentially none?_** 
- If we're able to restrict the field to just numbers, we do not need an error for other character entry (we should consider this for the zip code field as well)
- There is no specific minimum or maximum for this figure, but we could consider capping (though I don't know that it matters much if someone wishes to put 999999999, for example, as it just will be logged as "over 250k")

hamilton:
closes #496. We'll hold on this until (1) we can get the product owner to give a thumbs up and (2) we have a PR up for the schema.

A demo of this is viewable [here](https://vibrant-haibt-46f9d3.netlify.app/iframe.html?id=core-add-on-user-flows--demographics&viewMode=story).

Checklist for reviewer:


knowtheory (approved):

src/routes/demographics/Content.svelte
```diff
@@ -143,21 +166,61 @@
           {#if schema[question].type === 'text'}
             <div
               class="mzp-c-choice mzp-c-choice-text"
-              class:mzp-is-error={!$zipValidity.valid}>
+              class:mzp-is-error={
+                inputFormatters.showErrors(question) && // only show errors if explicitly told, e.g. on blur
+                inputFormatters.hasValidator(question) && // only show errors if there's even a validator
+                inputFormatters[question].isInvalid(results[question] // only show errors if the input is invalid
```
rhelmer, line 172: nit:
```suggestion
                inputFormatters[question].isInvalid(results[question]) // only show errors if the input is invalid
```

src/routes/demographics/Content.svelte
```diff
@@ -143,21 +166,61 @@
...
+            )}>
```
rhelmer, line 173: nit paired with the previous ;)
```suggestion
            }>
```

src/routes/demographics/Content.svelte
```diff
@@ -143,21 +166,61 @@
...
+                <span
+                  class="mzp-c-fieldnote"
+                  transition:fly={{ duration: 300, y: 5 }}>
+                  {inputFormatters[question].isInvalid(results[question])}
+                </span>
+              {/if}
+              </span>
               <!-- show validation errors here -->
-              <span style="min-height: 24px; display: block;">
+              <!-- <span style="min-height: 24px; display: block;">
```
rhelmer, line 215: Is there a reason to comment this out vs. just remove it? It's in the git history if you want it back... if you plan to do something with it like re-enable it when an issue is fixed please add a comment :)

src/routes/demographics/survey-tools.js
```diff
@@ -50,12 +50,21 @@ export function clearAnswer(questionType) {
 }
 
 export function zipcodeIsValid() {
-    return function zipcodeValid(value) {
-      if (Number.isNaN(Number(value)) || value.length > 5)
-        return "Please enter a five-digit zip code.";
-      else return true;
-    };
+  return function zipcodeValid(value) {
+    if (Number.isNaN(Number(value)) || value.length > 5)
```
rhelmer, line 54: Could you add braces here please? I thought we had an eslint rule for this...

src/routes/demographics/survey-tools.js
```diff
@@ -97,3 +106,92 @@ export function createFieldValidator(...validators) {
...
+      return value;
+    },
+    focus(event) {
+      const value = formatZipcodeInput(event);
+      this.showValidationErrors = false;
+      return value;
+    },
+    isInvalid(text) {
+      const number = Number(text);
+      if (text.length > 0 && text.length < 5) return 'zipcode must be a five-digit number';
```
rhelmer, line 169: I'm less worried about the missing brace if it's on the same line like this. The concern is just that someone will come along and add a line to it without realizing the missing brace (the cause of the "[goto fail](https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/)" bug, maybe most famously)

rhelmer (approved):
Just nits, looks OK to me. I'm concerned that we need to ratchet our eslint rules up a bit but I will tackle that in a separate issue (we already have one where vscode isn't aligning to what our CI is doing based on the same eslint config I'd like to figure out)

src/routes/demographics/Content.svelte
hamilton: hmm, this surfaces an ESLint error for me if I change this to what you're proposing. I'm going to just get rid of these comments.

src/routes/demographics/Content.svelte
hamilton: No good reason, removing now.

src/routes/demographics/MainDemographicsView.svelte
```diff
@@ -8,14 +8,22 @@ import { getContext, createEventDispatcher } from "svelte";
...
 // update the intermediate copy when $store.demographicsData changes.
+// This is additionally where we transform from the _response_ format -
+// what we will send to the server - to the _display_ format -
+// what is displayed to the user in the input fields.
 let intermediateResults;
-$: if ($store && $store.demographicsData) intermediateResults = { ...$store.demographicsData };
+$: if ($store && $store.demographicsData) {
+    intermediateResults = formatAnswersForDisplay(schema, { ...$store.demographicsData }, inputFormatters);
+}
+$: console.log(intermediateResults)
```
rhelmer, line 26: Did you mean to leave this?

tests/options-page/unit/formatters.test.js
```diff
@@ -0,0 +1,161 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
rhelmer, line 1: I am not necessarily opposed to it, but why should this live here vs. in the svelte tests?

src/routes/demographics/formatters.js
```diff
@@ -0,0 +1,243 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+ 
+ 
+ 
+ /** This Svelte action will format an input element on input, focus, and blur events.
```
rhelmer, line 7: Nit, but why does it only format on these events, and not as the user types? Wondering if that is a stylistic choice or just pragmatic. I can see the argument for not modifying input while the user is interacting.

rhelmer (approved):
The formatting functions themselves, plus tests, all lgtm. It also seems to work fine locally. I think some of the more Svelte-specific conventions are a little over my head still, but I like how generic it is.

src/routes/demographics/survey-schema.js
```diff
@@ -1,7 +1,9 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
```
rhelmer, line 1: Is there a PR up for the schema change ? It's nice to have it split out to its own JSON, how easily can we compare it to what's actually loaded on the server side?

I'm presuming that it is https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/main/schemas/pioneer-core/demographic-survey/demographic-survey.1.schema.json, in which case I guess it's not super directly comparable :) Maybe something to consider for the future.

src/routes/demographics/MainDemographicsView.svelte
hamilton: I sure didn't.

src/routes/demographics/survey-schema.js
hamilton: I agree. There is not a PR out and this schema is really just for the purposes of formatting (probably the wrong wording here). This part of the code base will probably be refactored when a real frontend engineer takes over this part of the codebase. It's not the most confusing setup ever, but it was slapped together pretty quickly and it'd be better to use an actual json schema approach or whatever.

src/routes/demographics/formatters.js
hamilton: I believe `input` is when it formats (which is when the user types / copy-pastes). @changecourse's requirements were to prevent inputs that shouldn't exist (like letters in a number field) instead of showing an error on incorrect input. I tried breaking the experience implemented here in various ways but was unsuccessful. This ultimately felt like the best way to control the nature of the inputs appropriate for (1) the form version of the answers (what should the entered-in format look like when the input field is not active, e.g. `$100,000` is what we see vs `100000` or 100000), (2) the edit version of the answers (we don't want the currency fields while typing to have the dollar sign and commas), and (3) the response version of the answers (what should the data look like when it goes to the servers?) These three things may all be different, hence this formatter.

src/routes/demographics/MainDemographicsView.svelte
rhelmer, line 26:
```suggestion
```

PR #558, opened 2021-04-26
Use remote schemaNamespace, not studyid

### LINKED ISSUE #557: enrollment and deletion pings should not use study addon ID as schema namespace
The core add-on is sending enrollment and deletion pings on the study's behalf, using the study add-on ID:

https://github.com/mozilla-rally/rally-core-addon/blob/fcc2375627b67af1132996a5db7a78bf55e48bed/core-addon/DataCollection.js#L121

This is essentially the same problem as https://github.com/mozilla-rally/rally-core-addon/pull/550 but the study ID is retrieved differently, it comes from https://github.com/mozilla-rally/rally-core-addon/blob/42b2a3810646557f4363bb6310efb34b06b89dc5/core-addon/Core.js#L184 which is the handler for `browser.management.onInstalled`.

If we do not want to use the study add-on ID as the namespace then we need to get this from the message that `rally.js` sends on initialization instead.

rhelmer:
fixes #557

Checklist for reviewer:

### DIFF
--- a/core-addon/Core.js
+++ b/core-addon/Core.js
@@ -289,6 +289,12 @@ export default class Core {
       throw new Error(`Core._handleExternalMessage - unexpected sender ${sender.id}`);
     }
 
+    const knownStudy = availableStudies.find(s => s.addonId == sender.id);
+    if (!("schemaNamespace" in knownStudy)) {
+      return Promise.reject(
+        new Error(`Core._handleExternalMessage - No schema namespace specified in remote settings for ${sender.id}`));
+    }
+
     let joinedStudies = await this._storage.getActivatedStudies();
     if (!joinedStudies.includes(sender.id)) {
       throw new Error(`Core._handleExternalMessage - ${sender.id} not joined`);
@@ -312,10 +318,11 @@ export default class Core {
         if (pausedStudies.includes(sender.id)) {
           throw new Error(`Core._handleExternalMessage - ${sender.id} is paused and may not send data`);
         }
-        const { payloadType, payload, namespace, keyId, key } = message.data;
+
+        const { payloadType, payload, keyId, key } = message.data;
         let rallyId = await this._storage.getRallyID();
         return await this._dataCollection.sendPing(
-          rallyId, payloadType, payload, namespace, keyId, key
+          rallyId, payloadType, payload, knownStudy.schemaNamespace, keyId, key
         );
       }
       default:
@@ -434,6 +441,11 @@ export default class Core {
       return Promise.reject(
         new Error(`Core._enrollStudy - Unknown study ${studyAddonId}`));
     }
+    const knownStudy = knownStudies.find(s => s.addonId == studyAddonId);
+    if (!("schemaNamespace" in knownStudy)) {
+      return Promise.reject(
+        new Error(`Core._enrollStudy - No schema namespace specified in remote settings for ${studyAddonId}`));
+    }
 
     // Record that user activated this study.
     await this._storage.appendActivatedStudy(studyAddonId);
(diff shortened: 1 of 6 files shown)


core-addon/Core.js
```diff
@@ -523,7 +542,12 @@ export default class Core {
 
       // Important: remove these lines once the migration
       // to Glean.js is finally complete.
-      await this._dataCollection.sendDeletionPing(rallyId, studyId);
+      const knownStudy = availableStudies.find(s => s.addonId == studyId);
```
rhelmer, line 545: One thing I am a little concerned with here: the list of studies that the user has ever been enrolled with is stored in `activatedStudies`, so unless we also store the matching `schemaNamespace` for each as well, then this basically means we cannot remove entries from remote-settings.

We could deal with migrating this data later, `activatedStudies` is currently an array of add-on IDs so we could make this could make that an array of object `{ studyId: schemaNamespace}` objects, or something along those lines.

We could make the remote-settings format much more efficient too, it's a JSON array-of-objects right now. Having an immutable list of past and present studies might not be the worst thing.

rhelmer: Hm actually I wonder if we need to do this at all, I filed https://github.com/mozilla-rally/rally-core-addon/issues/560 to discuss.

public/locally-available-studies.json
```diff
@@ -1,24 +1,31 @@
...
+    "studyDetailsLink": "https://rally.mozilla.org/current-studies/your-time-online-and-doomscrolling/",
+    "dataCollectionDetails": [
+      "Specific actions you take while browsing the web: loading a new URL, changing a tab, watching a video, or listening to audio (we do not collect the audio you are listening to, just that you have performed that action)",
+      "The domains you visit as you browse the web (e.g., wikipedia.org) and the title, description, and type of page that you're on (e.g., article, video, website)",
+      "The time spent on each page and how far you scroll down a page"
+    ],
+    "schemaNamespace": "rally-zero-one"
   }
-]
+]
```
Dexterp37, line 31: nit missing newline at the end :)

public/locally-available-studies.json
```diff
@@ -1,24 +1,31 @@
 [
   {
-    "name": "Rally Base Study",
+    "name": "Your Time Online and \"Doomscrolling\"",
```
Dexterp37, line 3: Mh, why are we changing this? It feels weird to do integration testing on a live study. This will end up sending bogus data .

Dexterp37 (changes requested):
My major concern over this is using a real study for integration testing. The other blocker for this is #560

public/locally-available-studies.json
rhelmer: @Dexterp37 It shouldn't send data because the core add-on will be in dev mode with sending data disabled, right? Happy to revert this.

rhelmer: I'm pretty sure we'll need to update the Basic Study (which is on AMO) to update `rally.js`, I was hoping we did not need to keep this "study" but if it's necessary then we can do so.

Maybe the real fix, though, is to use a synthetically-generated study for integration tests?

rhelmer: Just to be clear though our integration tests *shouldn't* be sending any data anywhere now, and I don't think changing this would change that.

I can actually go ahead and revert this, the reason we got here is a little complex but here it is for posterity:

1. we're maintaining this "Rally Base Study" on AMO just for testing purposes, while this has some benefit from a testing PoV it's using the network and takes a long time to update (add-on review)
2. I thought we'd need to update (1) but since I ended up using remote-settings it's not necessary after all
3. I don't think we should be sending data from the integration tests whether this is changed or not, that's configured by the core add-on

core-addon/Core.js
rhelmer: OK per #560 let's leave the current approach in place, and not remove the pings sent on uninstall.

rhelmer:
> My major concern over this is using a real study for integration testing. The other blocker for this is #560

OK these are both resolved.

public/locally-available-studies.json
Dexterp37: > Maybe the real fix, though, is to use a synthetically-generated study for integration tests?

Yes, the real fix is generating a study on the fly (we have the means to do that).

hamilton (approved):
lgtm, thanks for taking care of this @rhelmer !

Issue #545, opened 2021-04-15
Port the existing legacy telemetry data collection to Glean.js


Dexterp37:
The existing data collection is documented [here](https://github.com/mozilla-rally/rally-core-addon/blob/master/docs/DATA_COLLECTION.md). The pings:

- [x] enrollment (for the platform)
- [x] enrollment (for studies)
- [x] demographics
- [ ] deletion

Dexterp37:
@knowtheory @rhelmer 

Today I verified that we're ingesting data and that this is now ready to be turned on for collection. We will not turn off legacy telemetry collection from the Core Add-on just yet (studies still need to send data and we still need to validate).

However, we should:

- file a bug to create a new environment to send Glean.js collected data to for the Core Add-on.
- turn on Glean collection in the Core Add-on
- release a new version

What do you think, folks? Should we move forward?

knowtheory:
Yes please do move forward!

@Dexterp37 Do you need someone to file for the creation of a new env, or is that something you can tackle?

Dexterp37:
> 
> 
> Yes please do move forward!
> 
> @Dexterp37 Do you need someone to file for the creation of a new env, or is that something you can tackle?

This is done through [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1710406).

marniepw:
@Dexterp37 any updates here? thanks!

Dexterp37:
The deletion part is being handled in [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1698934) by Arkadiusz.

jepstein:
There is a pull request open to fix this

https://github.com/mozilla-rally/rally-core-addon/pull/696

jepstein:
pull request reviewed and merged

PR #577, opened 2021-05-04
Final touches to enable Glean.js

Dexterp37:
With the changes below, by providing the proper encryption key (the one currently being used for the legacy telemetry cannot be used for Glean.js, as we want to keep the environment separate), data collection with Glean.js in the core add-on is ready for prime time.

This contributes to #545 .

Checklist for reviewer:

### DIFF
--- a/core-addon/DataCollection.js
+++ b/core-addon/DataCollection.js
@@ -28,7 +28,7 @@ export default class DataCollection {
    *        Whether or not user has enrolled in the platform.
    */
   initialize(userEnrolled) {
-    if (!__ENABLE_GLEAN__) {
+    if (!__ENABLE_DATA_SUBMISSION__ || !__ENABLE_GLEAN__) {
       console.warn("DataCollection - Glean disabled by the build configuration.");
       return;
     }
--- a/core-addon/DataCollection.js
+++ b/core-addon/DataCollection.js
@@ -37,6 +37,7 @@ export default class DataCollection {
     // consented to join Rally. Upload is always enabled unless the web-extension
     // is uninstalled.
     Glean.initialize("rally-core", userEnrolled, {
+        appDisplayVersion: browser.runtime.getManifest().version,
         plugins: [
           new PingEncryptionPlugin(CORE_ENCRYPTION_JWK)
         ]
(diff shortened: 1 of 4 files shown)


core-addon/DataCollection.js
```diff
@@ -37,6 +37,7 @@ export default class DataCollection {
     // consented to join Rally. Upload is always enabled unless the web-extension
     // is uninstalled.
     Glean.initialize("rally-core", userEnrolled, {
+        appDisplayVersion: browser.runtime.getManifest().version,
```
rhelmer, line 40: I think this line caused a new unit test failure :( 
```
1) Core
       "before each" hook in "Core":
     Error: done() called multiple times in hook <Core "before each" hook in "Core"> of file /home/circleci/project/tests/core-addon/unit/Core.test.js; in addition, done() received error: TypeError: Cannot read property 'version' of undefined
    at DataCollection.initialize (file:///home/circleci/project/core-addon/DataCollection.js:40:57)
    at file:///home/circleci/project/core-addon/Core.js:38:28
    at <anonymous> {
  uncaught: true
}
      at process._fatalException (internal/process/execution.js:163:25)
      at processPromiseRejections (internal/process/promises.js:213:9)
      at processTicksAndRejections (internal/process/task_queues.js:94:32)
```

Specifically:
`TypeError: Cannot read property 'version' of undefined`

Is this all just mocked out in the unit tests? I know for integration we build a real `manifest.json` but just mocking out that call is probably just fine here.

rhelmer: I think `chrome.runtime.getManifest.returns({ version: FAKE_STUDY_VERSION })` should be enough, but I don't see a simple way to test that `FAKE_STUDY_VERSION` matches.

Maybe the version number could be validated and `DataCollection.js` could throw if it fails?

rhelmer (changes requested):

core-addon/DataCollection.js
Dexterp37: > Maybe the version number could be validated and `DataCollection.js` could throw if it fails?

We'd be testing the WebExtension APIs and the Glean APIs by doing this, since there's no specific code from the core add-on that does the collection.

Moreover, doing this kind of test would require us to inspect the Glean payload, which is a discouraged practice for a series of other reasons. I'd recommend leaving this as it is, since we do have tests in Glean.js that verify that configuration values provided at initialization make it into pings.

@rhelmer what do you think?

rhelmer (approved):

PR #601, opened 2021-05-17
Turn Glean data collection on

Dexterp37:
This makes sure Glean is turned on when packing the final add-on.

Let's not merge this until we give it a final QA pass :-)

Checklist for reviewer:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 [Full changelog](https://github.com/mozilla-rally/core-addon/compare/v1.2.1...master)
 
+* [#601](https://github.com/mozilla-rally/rally-core-addon/pull/601): Enable submitting data with Glean in addition to the legacy collection.
+
 # v1.2.1 (2021-04-27)
 
 [Full changelog](https://github.com/mozilla-rally/core-addon/compare/v1.2.0...v1.2.1)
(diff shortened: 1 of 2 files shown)


rhelmer (approved):
I did a non-dev-build locally and it seems to be working as expected - some pings are still using legacy Telemetry, but data from RS01 is attempting to use Glean.

@Dexterp37 do you have a more specific QA plan in mind? Can we check the analysis environment and ensure that it all looks OK? How does this intersect (if at all) with https://github.com/mozilla-rally/rally-study-01/pull/103 ?

Dexterp37:
> 
> 
> I did a non-dev-build locally and it seems to be working as expected - some pings are still using legacy Telemetry,

Yes, this is not disabling legacy telemetry, we're enabling them both to perform validation. 

> ... but data from RS01 is attempting to use Glean.

@rhelmer what do you mean here?

> @Dexterp37 do you have a more specific QA plan in mind? Can we check the analysis environment and ensure that it all looks OK? How does this intersect (if at all) with [mozilla-rally/rally-study-01#103](https://github.com/mozilla-rally/rally-study-01/pull/103) ?

The secure environment is already up (you were granted access too). I checked and pings are being ingested.
This is orthogonal to the RS01 study and should not have anything to do with it at all.

PR #771, opened 2022-01-11
bump rally-studies collection from -v1 to -v2, breaking compat, to en…

rhelmer:
…sure users have latest consent notices

Checklist for reviewer:

### DIFF
--- a/core-addon/Core.js
+++ b/core-addon/Core.js
@@ -755,6 +755,37 @@ export default class Core {
       return s;
     }).filter(study => (!study.studyPaused || study.studyJoined));
 
+    // Only show study if it is compatible with this version of the core add-on.
+    const coreVersion = browser.runtime.getManifest().version;
+
+    // Match the version number used by the core add-on, such as "1.3.7buildid20220107.052711".
+    // this uses capture groups to isolate the part before the build ID, the semantic version number "1.3.7".
+    const versionRegex = /^([0-9]+)\.([0-9]+)\.([0-9])s*/;
+    const [fullCore, majorCore, minorCore, patchCore] = coreVersion.match(versionRegex);
+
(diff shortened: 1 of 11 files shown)


knowtheory (approved):
lgtm

core-addon/Core.js
```diff
@@ -755,6 +755,37 @@ export default class Core {
       return s;
     }).filter(study => (!study.studyPaused || study.studyJoined));
 
+    // Only show study if it is compatible with this version of the core add-on.
+    const coreVersion = browser.runtime.getManifest().version;
+
+    // Match the version number used by the core add-on, such as "1.3.7buildid20220107.052711".
+    // this uses capture groups to isolate the part before the build ID, the semantic version number "1.3.7".
+    const versionRegex = /^([0-9]+)\.([0-9]+)\.([0-9])s*/;
```
knowtheory, line 763: what's with the `s*`?

public/studies-schema.json
```diff
@@ -87,6 +88,9 @@
     },
     "order": {
       "type": "number"
+    },
+    "minimumCoreVersion": {
+      "type": "number"
```
knowtheory, line 93: Is number the right type?  cause we're going to have things that are like `1.2.3` i would have thought the multiple `.` would require that we use string?

rhelmer: Good catch, thanks!

core-addon/Core.js
rhelmer: > what's with the `s*`?

The signing infra sets a long build ID as part of the version string, like: `1.3.7buildid20220107.052711`
I didn't want to have to worry about how long it might be, so I wanted to be on the safe side.

Any suggestions?

rhelmer: I'll make it `\\s*` otherwise

Issue #787, opened 2022-01-27
Core add-on update: turn off legacy telemetry for Glean.js studies


marniepw:
Related to: https://github.com/mozilla-rally/internal/issues/399

marniepw:
Rob found a bug; fixed and waiting on review from Carla and Mike.

rhelmer:
I've asked for approval from add-ons reviewers in https://mozilla.slack.com/archives/CMKP7NPKN/p1645030442650359

PR #788, opened 2022-02-01
issue #787 - ensure that data collection is always enabled in core ad…

rhelmer:
…d-on when user is enrolled

Checklist for reviewer:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 [Full changelog](https://github.com/mozilla-rally/core-addon/compare/v1.4.2...master)
 
+* [#788](https://github.com/mozilla-rally/rally-core-addon/pull/788): Ensure that data collection is always enabled in core add-on
+when user is enrolled
+
 # v1.4.2 (2022-01-26)
 
 [Full changelog](https://github.com/mozilla-rally/core-addon/compare/v1.4.1...v1.4.2)
--- a/core-addon/Core.js
+++ b/core-addon/Core.js
@@ -34,8 +34,12 @@ export default class Core {
 
     // Initialize the collection engine once we know if
     // user is enrolled or not.
-    this._storage.getRallyID().finally(id =>
-      this._dataCollection.initialize(id !== undefined));
+    this._storage.getRallyID().then(id => {
+      this._dataCollection.initialize(id !== undefined)
+    }).catch(err => {
+      console.error("No Rally ID, initializing data collection in unenrolled state", err);
+      this._dataCollection.initialize(false);
+    });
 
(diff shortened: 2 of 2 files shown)


rhelmer:
I noticed this when testing https://github.com/mozilla-rally/rally-core-addon/issues/787 - data collection will not always be enabled for Glean in the core add-on after a restart.

I believe it is because this code uses `Promise.finally()`, but expecting an argument which [according to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally#description) will not be provided:

> A finally callback will not receive any argument, since there's no reliable means of determining if the promise was fulfilled or rejected. This use case is for precisely when you do not care about the rejection reason, or the fulfillment value, and so there's no need to provide it.

I think this is fortunately not a huge deal, yet, because:

1. glean-based studies send their own enrollment ping
2. glean will send deletion pings regardless

I believe that `finally()` was used here because the intent was to initialize data collection regardless of whether the user was enrolled at the time it was initialized, which I think is the correct thing to do (for important functions like the glean deletion ping mentioned above) so I've preserved that behavior with `then` and `catch`.

rhelmer:
> I noticed this when testing #787 - data collection will not always be enabled for Glean in the core add-on after a restart.
> 
> I believe it is because this code uses `Promise.finally()`, but expecting an argument which [according to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally#description) will not be provided:
> 
> > A finally callback will not receive any argument, since there's no reliable means of determining if the promise was fulfilled or rejected. This use case is for precisely when you do not care about the rejection reason, or the fulfillment value, and so there's no need to provide it.
> 
> I think this is fortunately not a huge deal, yet, because:
> 
>     1. glean-based studies send their own enrollment ping
> 
>     2. glean will send deletion pings regardless
> 
> 
> I believe that `finally()` was used here because the intent was to initialize data collection regardless of whether the user was enrolled at the time it was initialized, which I think is the correct thing to do (for important functions like the glean deletion ping mentioned above) so I've preserved that behavior with `then` and `catch`.

All that said, I think we should take this patch in the very next release, because:

1. we are turning off legacy telemetry in #787
2. it makes it much easier to QA this

Carla-Moz (approved):
Thanks for the reviewer checklist and explanations; quite helpful. Built and tested, lgtm!

stansky (approved):

rhelmer:
Thanks! I made a small change to the error message that is logged that makes it clear that data collection is initialized, but that the user is unenrolled. Glean.js should log additional info after that, so it'll be much clearer why this is happening, I hope.

PR #881, opened 2023-04-06
Decommisioning extension

thomik-corp:
Decommissioning the core addon.

### DIFF
--- a/core-addon/Core.js
+++ b/core-addon/Core.js
@@ -85,6 +85,12 @@ export default class Core {
     browser.browserAction.onClicked.addListener(this._openControlPanel);
     // After installing the addon, make sure to show the control page.
     browser.runtime.onInstalled.addListener(async ({ reason, temporary }) => {
+      // If core-addon is to be revived, then remove lines starting here...
+      console.log("Mozilla Core AddOn has been decommissioned. Uninstalling self...");
+      browser.management.uninstallSelf();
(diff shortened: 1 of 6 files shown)


core-addon/Core.js
```diff
@@ -85,6 +85,12 @@ export default class Core {
     browser.browserAction.onClicked.addListener(this._openControlPanel);
     // After installing the addon, make sure to show the control page.
     browser.runtime.onInstalled.addListener(async ({ reason, temporary }) => {
+      // If core-addon is to be revived, then remove lines starting here...
```
Dexterp37, line 88: I believe we should be clearing the storage as well and potentially uninstalling all the studies. e.g. do what [this](https://github.com/mozilla-rally/rally-core-addon/blob/1bf919f595d5ef4d1a4b2be908cecd3bcf2d6974/core-addon/Core.js#L547) is doing.

Or is there a different plan?

aaga: We actually don't want to do "unenroll" as that will send data deletion requests – and we have some customers that are still doing analysis. We also don't need to uninstall studies because we are pushing "self-uninstalling" updates for all studies.

aaga: (we preferred to have studies uninstall themselves rather than via Core Add-on, in case some people had studies installed without Core Add-on)

